PR #12, opened 2021-10-04
Enrich graph with monitoring

oravi:


lineage/lineage_graph.py
```diff
@@ -249,7 +273,42 @@ def _update_selected_node_attributes(self, selected_node: str) -> None:
             node.update({'color': self.SELECTED_NODE_COLOR,
                          'title': self.SELECTED_NODE_TITLE + node_title})
 
+    def _get_freshness_and_volume_graph_for_node(self, node: str) -> str:
```
marcellovictorino, line 273: I would recommend decoupling this from the main `lineage_graph.py` logic.
It does not really have to know how the graph is being generated, as long as it receives the `html div block with the expected style, title, and img.

Personally, I find matplotlib charts quite outdated, requiring a lot of imperative effort to change and look nice.
I would recommend looking into Plotly or Altair (my favourite), which are also capable of providing interactive viz 👍

lineage/lineage_graph.py
```diff
@@ -302,6 +303,22 @@ def _enrich_graph_with_monitoring_context(self) -> None:
                     </body>
                 </html>
                 """
+
+                node_volumes = self.catalog[node]['volume']
+                if node_volumes[-1] < median(node_volumes) / 2:
```
marcellovictorino, line 308: Nice!
Similarly, possible to use Strategy Pattern here as well. Decoupling the "anomaly detection" implementation from the display itself

I would consider using the [Tukey's fence method](https://en.wikipedia.org/wiki/Outlier#Tukey's_fences), where the value is greater/lesser than 1.5 x the Inter Quartile Range

PR #78, opened 2022-05-31
DbtRunner's run_operation also logs errors.

elongl:
Currently, `DbtRunner.run_operation()` ignores whether running the macro succeeded or failed and solely extracts Elementary's logs.
This is a bit unfortunate as the output contains useful information to the user, or potential contributor, regarding what went wrong, which is one of the main point of tests.
I've updated the code that checks whether the operation failed, and if it did, log the errors.

### DIFF
--- a/monitor/dbt_runner.py
+++ b/monitor/dbt_runner.py
@@ -66,7 +66,7 @@ class DbtRunner(object):
                 log_message_dict = json.loads(json_message)
                 log_message_data_dict = log_message_dict.get('data')
                 if log_message_data_dict is not None:
-                    if not success and log_message_dict['level'] == 'error':
+                    if log_message_dict['level'] == 'error':
                         logger.error(log_message_data_dict)
                         continue
                     log_message = log_message_data_dict.get('msg')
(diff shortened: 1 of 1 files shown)


monitor/dbt_runner.py
```diff
@@ -56,13 +57,18 @@ def run_operation(self, macro_name: str, json_logs: bool = True, macro_args: dic
...
+        if not success:
+            logger.error(f'Failed to run macro: "{macro_name}"')
         run_operation_results = []
         if json_logs:
             json_messages = command_output.splitlines()
             for json_message in json_messages:
                 log_message_dict = json.loads(json_message)
                 log_message_data_dict = log_message_dict.get('data')
                 if log_message_data_dict is not None:
+                    if log_message_dict['level'] == 'error':
```
oravi, line 69: Are you sure 'level' always exists in log_message_dict? if you are not sure use 'get' 
Also, let's introduce this change more gradually so please add a param to the func called 'log_errors' that is disabled by default. Then in the unit-tests we will pass True there but in our users' deployments it won't be enabled for now by default.

elongl: Yes, I'm confident that the `level` property is always present.  
The [`LogMessage`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/logger.py#L50) class has a mandatory `level` field. I'll add the `logs_errors`.

oravi: I'd still use get as dbt might change that in the future but I'll merge the PR and edit it before releasing the next version, thanks!

elongl: There it is - [PR](https://github.com/elementary-data/dbt-data-reliability/pull/29).

PR #79, opened 2022-06-08
Add support for vars argument in dbt runner methods

IDoneShaveIt:
Adding a support for passing vars to the following dbt runner methods:

1. run_operation
2. run
3. test

### DIFF
--- a/monitor/dbt_runner.py
+++ b/monitor/dbt_runner.py
@@ -33,7 +33,7 @@ class DbtRunner(object):
             dbt_command.extend(['--target', self.target])
         if vars:
             json_vars = json.dumps(vars)
-            command_args.extend(['--vars', json_vars])
+            dbt_command.extend(['--vars', json_vars])
         logger.info(f"Running {' '.join(dbt_command)} (this might take a while)")
         result = subprocess.run(dbt_command, check=False, capture_output=capture_output)
         output = None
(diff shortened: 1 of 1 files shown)


monitor/dbt_runner.py
```diff
@@ -51,12 +51,21 @@ def snapshot(self) -> bool:
...
 
-    def run_operation(self, macro_name: str, json_logs: bool = True, macro_args: dict = None,
-                      log_errors: bool = False) -> list:
+    def run_operation(
+        self,
+        macro_name: str,
+        json_logs: bool = True,
+        macro_args: Optional[dict] = None,
+        log_errors: bool = False,
+        dbt_vars: Optional[dict] = None
```
oravi, line 60: cosmetics -  I'd rename dbt_vars to vars as it's already inside DbtRunner so no need for the prefix dbt here

monitor/dbt_runner.py
```diff
@@ -51,12 +51,21 @@ def snapshot(self) -> bool:
...
+    ) -> list:
         command_args = ['run-operation', macro_name]
         if macro_args is not None:
             json_args = json.dumps(macro_args)
             command_args.extend(['--args', json_args])
+        if dbt_vars is not None:
```
oravi, line 66: Since vars is supported by all dbt commands, I'd implement this code inside _run_command to avoid the code duplication

PR #83, opened 2022-06-13
SlackClient that uses slack token and supports sending messages and uploading files

IDoneShaveIt:
- Created SlackClient that support both slack webhook and slack token
- Support upload files to slack using SlackClient with slack token only (webhook does not support uploading files)
- Replaced the use of the old slack messages send method with the new SlackClient
- Fixed a bug where we marked alerts as sent even if they failed to


requirements.txt
```diff
@@ -19,3 +19,4 @@ beautifulsoup4
 posthog
 sqlfluff
 ruamel.yaml
+slack_sdk
```
IDoneShaveIt, line 22: not sure if needed to be added manually

oravi (changes requested):

requirements.txt
oravi: it's generated automatically but the result will be the same so it's ok :)

clients/slack/schema.py
```diff
@@ -0,0 +1,8 @@
+from typing import Optional
+from pydantic import BaseModel
+
+
+class SlackMessageSchema(BaseModel):
+    text: Optional[str] = None
+    attachments: Optional[list] = None
```
IDoneShaveIt, line 7: In the future we will create schemas for the attachments and the blocks as well

oravi: The only concern I have with it, is that it's a new requirement and if it's not in use right now I'd rather avoid adding requirements. Also, if you decide to keep it - add it to the requirements and setup files

monitor/cli.py
```diff
@@ -145,5 +179,87 @@ def report(ctx, config_dir, profiles_dir, update_dbt_package, profile_target):
...
+    track_cli_start(anonymous_tracking, 'monitor-send-report', get_cli_properties(), ctx.command.name)
+    try:
+        data_monitoring = DataMonitoring(
+            config=config,
+            force_update_dbt_package=update_dbt_package,
+            slack_webhook=slack_webhook,
+            slack_token=slack_token,
+            slack_channel_name=slack_channel_name
+        )
+        generated_report_successfully = data_monitoring.generate_report()
```
oravi, line 254: cosmetics - send report can call generate_report internally

monitor/data_monitoring.py
```diff
@@ -26,20 +28,30 @@ class DataMonitoring(object):
...
+        slack_channel_name: Optional[str] = None
+    ) -> None:
         self.config = config
         self.dbt_runner = DbtRunner(self.DBT_PROJECT_PATH, self.config.profiles_dir, self.config.profile_target)
         self.execution_properties = {}
         self.slack_webhook = slack_webhook or self.config.slack_notification_webhook
+        self.slack_token = slack_token or self.config.slack_token
+        self.slack_channel_name = slack_channel_name or self.config.slack_notification_channel_name
+        # slack client is optional
+        self.slack_client = SlackClient.init(token=self.slack_token, webhook=self.slack_webhook) if (self.slack_token or self.slack_webhook) else None
```
oravi, line 46: Why you don't provide the channel_name here to init as well? also - looking at the code and I was thinking that maybe init should be 'create_slack_client' as it's basically a factory method

monitor/data_monitoring.py
```diff
@@ -152,6 +169,21 @@ def generate_report(self) -> bool:
             self.execution_properties['report_end'] = True
             self.execution_properties['success'] = self.success
             return self.success
+    
+    def send_report(self) -> bool:
+        elementary_html_file_path = os.path.join(self.config.target_dir, 'elementary.html')
```
oravi, line 174: I'd call here the generate_report. Also, generate_report could return back elementary_html_file_path (in addition to the success flag). This way we could add to the path also the report creation time (we already add it to the json but it will be good to have it in the report name). WDYT?

monitor/data_monitoring.py
```diff
@@ -152,6 +169,21 @@ def generate_report(self) -> bool:
...
+        if os.path.exists(elementary_html_file_path):
+            file_uploaded_succesfully = self.slack_client.upload_file(
+                channel_name=self.slack_channel_name,
+                file_path=elementary_html_file_path,
+                message=SlackMessageSchema(text="Elementary monitoring report")
+            )
+            if not file_uploaded_succesfully:
+                self.success = False
+        else:
+            logger.error('Could not send "Elementary monitor report" because it is not exists.')
```
oravi, line 184: cosmetics - 'Could not send Elementary monitoring report because it does not exist'

setup.py
```diff
@@ -31,6 +31,8 @@
         'posthog',
         'ruamel.yaml',
         'alive-progress<=2.3.1'
+        'slack_sdk',
```
oravi, line 34: Maybe let's limit these to their current major versions? for example slack_sdk < 4.0

clients/slack/schema.py
IDoneShaveIt: Well I do use it :) 
Just not implemented the attachments schema all the way down, but we now have a schema for slack message which I think is good and useful :)

monitor/data_monitoring.py
IDoneShaveIt: Will change the name :)

monitor/cli.py
IDoneShaveIt: Decided to separate between sending and generating

PR #89, opened 2022-06-20
Fixed failure on singular tests

oravi:


monitor/cli.py
```diff
@@ -132,12 +132,12 @@ def report(ctx, config_dir, profiles_dir, update_dbt_package, profile_target):
...
     try:
         data_monitoring = DataMonitoring(config, update_dbt_package)
-        data_monitoring.generate_report()
+        status_code = data_monitoring.generate_report()
         track_cli_end(anonymous_tracking, 'monitor-report', data_monitoring.properties(), ctx.command.name)
     except Exception as exc:
         track_cli_exception(anonymous_tracking, 'monitor-report', exc, ctx.command.name)
         raise
-    return
+    return status_code
```
IDoneShaveIt, line 140: What does the return status_code used for?

monitor/data_monitoring.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, config: Config, force_update_dbt_package: bool = False,
         self.execution_properties = {}
         self.slack_webhook = slack_webhook or self.config.slack_notification_webhook
         self._download_dbt_package_if_needed(force_update_dbt_package)
+        self.status_code = 0
```
IDoneShaveIt, line 33: Well I like being more specific.
status code is used for? 0 and 1 are numbers with no real meaning.
IMO it is better to call it run_successfully and set it t to True/False.
It much more explicit :) 
WDYT?

monitor/data_monitoring.py
```diff
@@ -93,18 +101,22 @@ def _send_alerts(self, days_back: int):
...
+    def run(self, days_back: int, dbt_full_refresh: bool = False) -> int:
 
         logger.info("Running internal dbt run to aggregate alerts")
         success = self.dbt_runner.run(models='alerts', full_refresh=dbt_full_refresh)
         self.execution_properties['alerts_run_success'] = success
         if not success:
             logger.info('Could not aggregate alerts successfully')
-            return
+            self.execution_properties['run_status_code'] = 1
+            return 1
```
IDoneShaveIt, line 112: return self.status_code

monitor/data_monitoring.py
```diff
@@ -93,18 +101,22 @@ def _send_alerts(self, days_back: int):
...
-    def run(self, days_back: int, dbt_full_refresh: bool = False) -> None:
+    def run(self, days_back: int, dbt_full_refresh: bool = False) -> int:
 
         logger.info("Running internal dbt run to aggregate alerts")
         success = self.dbt_runner.run(models='alerts', full_refresh=dbt_full_refresh)
         self.execution_properties['alerts_run_success'] = success
         if not success:
             logger.info('Could not aggregate alerts successfully')
-            return
+            self.execution_properties['run_status_code'] = 1
```
IDoneShaveIt, line 111: In case that run will not be the last thing that can change the status_code, it is better to set the status_code and pass it explicit to the execution_properties
```suggestion
            self.status_code = 1
            self.execution_properties['run_status_code'] = self.status_code
```

monitor/test_result.py
```diff
@@ -15,12 +17,16 @@ def __init__(self, id, model_unique_id, test_unique_id, status) -> None:
...
-        else:
-            return ElementaryTestResult(**test_result_dict)
+        try:
+            if test_type == 'dbt_test':
+                return DbtTestResult(**test_result_dict)
+            else:
+                return ElementaryTestResult(**test_result_dict)
+        except Exception:
+            logger.error(f"Failed parsing test result - {json.dumps(test_result_dict)}")
+        return None
```
IDoneShaveIt, line 29: IMO its clearer that the return None is due to the exception if its inside the exception part :)

monitor/test_result.py
```diff
@@ -15,12 +17,16 @@ def __init__(self, id, model_unique_id, test_unique_id, status) -> None:
...
     @staticmethod
-    def create_test_result_from_dict(test_result_dict: dict) -> 'TestResult':
+    def create_test_result_from_dict(test_result_dict: dict) -> Optional['TestResult']:
         test_type = test_result_dict.get('test_type')
-        if test_type == 'dbt_test':
-            return DbtTestResult(**test_result_dict)
-        else:
-            return ElementaryTestResult(**test_result_dict)
+        try:
+            if test_type == 'dbt_test':
```
IDoneShaveIt, line 23: I know that its a moving of a code, but I think that strings should be consts.
In that case because we have a close amount of options, even an Enum :) 

but this is less important for now.

monitor/test_result.py
```diff
@@ -62,7 +68,7 @@ def send_to_slack(self, webhook: str, is_slack_workflow: bool = False):
         requests.post(url=webhook, headers={'Content-type': "application/json"}, data=json.dumps(data))
 
     @staticmethod
-    def display_name(str_value):
+    def display_name(str_value: Optional[str]) -> str:
```
IDoneShaveIt: If typing is Optional, then we need to pass default value of None.

PR #99, opened 2022-07-03
Tests executions api

IDoneShaveIt:
Extended the report to include "Tests Runs" screen data:

1. Improved code structure and created "APIs" for fetching data for the report screens
2. Made the code less script, and split the code into reusable methods
3. Generate additional data for "Tests Runs" screen (test runs, coverage)


monitor/api/models/models.py
```diff
@@ -0,0 +1,95 @@
...
+        normalized_model['model_name'] = model_name
+        normalized_model['normalized_full_path'] = ModelsAPI._normalize_model_path(
+            model_path=model.full_path,
+            model_package_name=model.package_name,
+            is_source=is_source
+        )
+        return NormalizedModelSchema(**normalized_model)
+    
+    @staticmethod
+    def _normalize_model_path(model_path: str, model_package_name: Optional[str] = None, is_source: bool = False) -> str:
```
IDoneShaveIt, line 79: moved from `data_monitoring` to here

monitor/api/models/models.py
```diff
@@ -0,0 +1,95 @@
...
+        if coverage_results:
+            for coverage_result in json.loads(coverage_results[0]):
+                coverages[coverage_result["model_unique_id"]] = ModelCoverageSchema(
+                    table_tests=coverage_result["table_tests"],
+                    column_tests=coverage_result["column_tests"]
+                )
+        return coverages
+
+    @staticmethod
+    def _normalize_dbt_model_dict(model: ModelSchema, is_source: bool = False) -> NormalizedModelSchema:
```
IDoneShaveIt, line 48: moved from `data_monitoring` to here

clients/api/api.py
```diff
@@ -0,0 +1,10 @@
+from clients.dbt.dbt_runner import DbtRunner
+
+
+class APIClient:
```
IDoneShaveIt, line 4: In the future we will add a wrapper that adds error/successful status for each call 🙂

monitor/api/models/schema.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class NormalizedModelSchema(ModelSchema):
+    owners: Optional[List[str]] = []
+    tags: Optional[List[str]] = []
+    model_name: str
+    normalized_full_path: str
+
+    @validator("owners", pre=True, always=True)
+    def set_owners(cls, owners):
```
IDoneShaveIt, line 25: This is the way to set an empty list/dict for a default value

monitor/api/models/schema.py
```diff
@@ -0,0 +1,35 @@
...
+        return owners or []
+
+    @validator("tags", pre=True, always=True)
+    def set_tags(cls, tags):
```
IDoneShaveIt, line 29: This is the way to set an empty list/dict for a default value

monitor/api/sidebar/sidebar.py
```diff
@@ -0,0 +1,49 @@
...
+        for model in [*models.values(), *sources.values()]:
+            SidebarAPI._update_dbt_sidebar(
+                dbt_sidebar=sidebar,
+                model_unique_id=model.unique_id,
+                model_full_path=model.normalized_full_path
+            )
+        return sidebar
+
+    @staticmethod
+    def _update_dbt_sidebar(dbt_sidebar: dict, model_unique_id: str, model_full_path: str) -> None:
```
IDoneShaveIt, line 35: moved from `data_monitoring` to here

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,165 @@
+import json
+from typing import Any, Dict, List, Optional
+
+from clients.api.api import APIClient
+from monitor.api.tests.schema import TestMetadataSchema, TestUniqueIdType, InvocationSchema
+
+
+class TestsAPI(APIClient):
+    @staticmethod
+    def _get_test_sub_type_unique_id(test: TestMetadataSchema) -> str:
```
IDoneShaveIt, line 10: This is a temp method until we'll split the all column/table tests into units

monitor/dbt_project/macros/get_test_results.sql
```diff
@@ -24,19 +24,6 @@
...
-
-        {% set test_rows_sample = none %}
-        {%- if status != 'error'-%}
-            {% set test_rows_sample = elementary_internal.get_test_rows_sample(test_results_query, test_type, results_sample_limit) %}
-        {%- endif -%}
-        {% do test_result_dict.update({'test_rows_sample': test_rows_sample}) %}
-        {% do test_result_with_sample_dicts.append(test_result_dict) %}
-    {%- endfor -%}
-    {% do elementary.edr_log(tojson(test_result_with_sample_dicts)) %}
+    {% do elementary.edr_log(tojson(test_result_dicts)) %}
```
IDoneShaveIt, line 27: moved this inner loop to a different macro and call it by itself

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+        self,
+        tests_metadata: Optional[List[TestMetadataSchema]] = None,
+        days_back: Optional[int] = None,
+        metrics_sample_limit: int = 5
+    ) -> Dict[TestUniqueIdType, Dict[str, Any]]:
+        tests: List[TestMetadataSchema] = tests_metadata if tests_metadata is not None else self.get_tests_metadata(days_back=days_back)
+        tests = [dict(test) for test in tests]
+        return json.loads(self.dbt_runner.run_operation(
+            macro_name='get_tests_metrics',
+            macro_args=dict(tests=tests, metrics_sample_limit=metrics_sample_limit)
```
oravi, line 27: This could be an issue actually - CLI has a limit in terms of the length of the params (this is why we split the slack ids to batches). Let's discuss about it?

monitor/dbt_project/macros/get_tests_metrics.sql
```diff
@@ -0,0 +1,23 @@
+{% macro get_tests_metrics(tests, metrics_sample_limit = 5) %}
```
oravi, line 1: See comment above about the limitation of the CLI

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def get_metrics(
+        self,
+        tests_metadata: Optional[List[TestMetadataSchema]] = None,
+        days_back: Optional[int] = None,
+        metrics_sample_limit: int = 5
+    ) -> Dict[TestUniqueIdType, Dict[str, Any]]:
+        tests: List[TestMetadataSchema] = tests_metadata if tests_metadata is not None else self.get_tests_metadata(days_back=days_back)
+        tests = [dict(test) for test in tests]
+        return json.loads(self.dbt_runner.run_operation(
```
oravi, line 25: Validate result before loading? also validate that you have one item?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+class TestsAPI(APIClient):
+    @staticmethod
+    def _get_test_sub_type_unique_id(test: TestMetadataSchema) -> str:
+        return f"{test.test_unique_id}.{test.test_type}.{test.test_sub_type if test.test_sub_type else 'None'}.{test.table_name if test.table_name else 'None'}.{test.column_name if test.column_name else 'None'}"
+
+    def get_tests_metadata(self, days_back: Optional[int] = 7) -> List[TestMetadataSchema]:
+        tests_metadata = json.loads(self.dbt_runner.run_operation(macro_name='get_test_results', macro_args=dict(days_back=days_back))[0])
+        return [TestMetadataSchema(**test_metadata) for test_metadata in tests_metadata]
+
+    def get_metrics(
```
oravi, line 17: cosmetics - maybe rename to get_tests_sample_data?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+        )[0])
+
+    def get_invocations(
+        self,
+        invocations_per_test: int = 30
+    ) -> Dict[TestUniqueIdType, InvocationsSchems]:
+        tests_invocations = dict()
+        raw_invocations = json.loads(self.dbt_runner.run_operation(
+            macro_name='get_tests_invocations',
+            macro_args=dict(invocations_per_test=invocations_per_test)
```
oravi, line 37: This seems like you pull the test invocations for the entire 'elementary_test_results' table which will grow over time. You should use / limit by the days_back (see the code in get_test_results - maybe even use it?)

monitor/dbt_project/macros/get_tests_invocations.sql
```diff
@@ -0,0 +1,63 @@
...
+        group by 
+            test_unique_id,
+            test_type,
+            test_sub_type,
+            table_name,
+            column_name
+    {% endset %}
+    {% set tests_invocations_agate = run_query(tests_invocations_query) %}
+    {% set tests_invocations_results = elementary.agate_to_dicts(tests_invocations_agate) %}
+    {% for test in tests_invocations_results %}
```
oravi, line 52: As you already iterate over the results in the python function to unite the different arrays - maybe do this code there instead of going over the results again?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+        )[0])
+        for sub_test_unique_id, invocations in raw_invocations.items():
+            test_invocations = []
+            test_invocations_times = json.loads(invocations["invocations_times"])
```
oravi, line 41: I think that if you use array_agg you won't need this json load again

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+            test_invocations_statuses = json.loads(invocations["statuses"])
+            for index, invocation_id in enumerate(test_invocations_ids):
+                test_invocations.append(InvocationSchema(
+                    id=invocation_id,
+                    time_utc=test_invocations_times[index],
+                    status=test_invocations_statuses[index],
+                    affected_rows=self._parse_affected_row(affected_rows=test_invocations_affected_rows, index=index)
+                ))
+
+            totals = self._get_invocations_totals(test_invocations)
```
oravi, line 53: cosmetics - _get_test_invocations_totals?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+            resolve=0
+        )
+
+    @staticmethod
+    def _get_invocations_description(invocations_totals: TotalsInvocationsSchema) -> str:
+        all_invocations_count = invocations_totals.errors + invocations_totals.warnings + invocations_totals.passed + invocations_totals.resolve
+        return f"There were {invocations_totals.errors or 'no'} failures and {invocations_totals.warnings or 'no'} warnings out of the last {all_invocations_count} test runs."
+
+    @staticmethod
+    def _parse_affected_row(affected_rows: List[str], index: int) -> Optional[int]:
```
oravi, line 80: I think that if we will extract in the query using a regex we won't need this

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+        try:
+            return int(affected_rows[index])
+        except Exception:
+            return None
+
+    def get_total_tests_results(self, tests_metadata: Optional[List[TestMetadataSchema]] = None, days_back: Optional[int] = None):
```
oravi, line 86: cosmetics - maybe rename to get_totals_for_test_results_screen?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+        totals = dict()
+        for test in tests:
+            self._update_test_results_totals(
+                totals_dict=totals,
+                model_unique_id=test.model_unique_id,
+                status=test.status
+            )
+        return totals
+    
+    def get_total_tests_runs(
```
oravi, line 97: cosmetics - maybe rename to get_totals_for_test_runs_screen?

monitor/data_monitoring.py
```diff
@@ -187,147 +196,68 @@ def send_report(self, elementary_html_path: str) -> bool:
...
+        serializeable_models = dict()
+        for key in models_and_sources.keys():
+            serializeable_models[key] = dict(models_and_sources[key])
+
+        dbt_sidebar = sidebar_api.get_sidebar(models=models, sources=sources)
         
-        return os.path.sep.join(splited_model_path)
+        return serializeable_models, dbt_sidebar
+    
+    def _get_dbt_models_coverages(self) -> Dict[str, Dict[str, int]]:
```
oravi, line 255: get_dbt_models_test_coverage?

monitor/dbt_project/macros/get_coverages.sql
```diff
@@ -0,0 +1,33 @@
+{% macro get_coverages() %}
```
oravi, line 1: get_dbt_models_test_coverage?

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,163 @@
...
+    def get_invocations(
+        self,
+        invocations_per_test: int = 30
+    ) -> Dict[TestUniqueIdType, InvocationsSchems]:
+        tests_invocations = dict()
+        raw_invocations = json.loads(self.dbt_runner.run_operation(
+            macro_name='get_tests_invocations',
+            macro_args=dict(invocations_per_test=invocations_per_test)
+        )[0])
+        for sub_test_unique_id, invocations in raw_invocations.items():
```
oravi, line 39: Add try-except to avoid failing if the parsing for a specific test failed?

monitor/api/models/models.py
```diff
@@ -0,0 +1,95 @@
...
+
+        tags = model.tags
+        if tags:
+            loaded_tags = try_load_json(tags)
+            if loaded_tags is not None:
+                tags = loaded_tags
+            else:
+                tags = [tags]
+        
+        normalized_model = dict(model)
```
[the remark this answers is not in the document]
IDoneShaveIt: I will parse it into json and then load it for creating a deep copy.
I am really against changing the original object to prevent weird behaviours if used in several places.
Plus we are creating a normalized model which is a different thing :) 
(In the future we can pass the normalization into the schema itself)

monitor/api/models/models.py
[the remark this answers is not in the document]
IDoneShaveIt: no need for the cls object here but sure I don't mind :)

monitor/api/models/models.py
```diff
@@ -0,0 +1,95 @@
+import json
```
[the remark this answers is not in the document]
IDoneShaveIt: Its under api directory so I thought its redundant 🤷🏻‍♂️

monitor/api/sidebar/sidebar.py
```diff
@@ -0,0 +1,49 @@
...
+        self.models_api = ModelsAPI(dbt_runner=self.dbt_runner)
+        self.tests_api = TestsAPI(dbt_runner=self.dbt_runner)
+
+    def get_sidebar(
+        self,
+        models: Optional[Dict[str, NormalizedModelSchema]] = None,
+        sources: Optional[Dict[str, NormalizedModelSchema]] = None
+    ) -> dict:
+        sidebar = dict()
+        models = models if models is not None else self.models_api.get_models()
```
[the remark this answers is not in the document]
IDoneShaveIt: In the long run this API can be used by different features.
Some might need the sidebar only, and some won't.
I thought its a good practice to make it optional, and query the models/sources only if needed.
I can change that though

monitor/api/tests/schema.py
```diff
@@ -0,0 +1,50 @@
...
+    time_utc: str
+    id: str
+    status: str
+
+
+class TotalsInvocationsSchema(BaseModel):
+    errors: Optional[int] = 0
+    warnings: Optional[int] = 0
+    passed: Optional[int] = 0
+    resolve: Optional[int] = 0
```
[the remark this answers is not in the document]
IDoneShaveIt: It was resolve when I encountered it.
I can change it

monitor/dbt_project/macros/get_sub_test_unique_id.sql
```diff
@@ -0,0 +1,4 @@
+{% macro get_sub_test_unique_id(test_unique_id, test_type, test_sub_type, table_name, column_name) %}
+    {% set sub_test_unique_id = test_unique_id ~ '.' ~ test_type ~ '.' ~ test_sub_type ~ '.' ~ table_name ~ '.' ~ column_name %}
```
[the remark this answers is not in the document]
IDoneShaveIt: The report uses:
```python
f"{test.test_unique_id}.{test.test_type}.{test.test_sub_type if test.test_sub_type else 'None'}.{test.table_name if test.table_name else 'None'}.{test.column_name if test.column_name else 'None'}"
```
which is the same as the macro :)

monitor/dbt_project/macros/get_tests_invocations.sql
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        select 
+            test_unique_id,
+            test_type,
+            test_sub_type,
+            table_name,
+            column_name,
+            SPLIT(
+               {{ dbt_utils.listagg(measure='detected_at', delimiter_text="';;;'", order_by_clause="order by detected_at", limit_num=invocations_per_test)}},
```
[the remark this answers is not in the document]
IDoneShaveIt: Writing here for documentation:
array_agg and array_slice are not supported on all platforms, therefore using listagg is the right thing to do.
About the desc order - the frontend expect it to be in that order 🤷🏻‍♂️

monitor/dbt_project/macros/get_tests_invocations.sql
IDoneShaveIt: requires to many changes for now,  can be an improvements in the next prs

monitor/api/tests/tests.py
IDoneShaveIt: Can't use array_agg because it is not supported by all platforms (as mentioned before)

monitor/api/tests/tests.py
IDoneShaveIt: I don't like cupeling between BE and FE

monitor/api/tests/tests.py
IDoneShaveIt: I don't like cupeling between BE and FE

monitor/api/tests/tests.py
```diff
@@ -0,0 +1,183 @@
...
+
+                totals = self._get_test_invocations_totals(test_invocations)
+                tests_invocations[sub_test_unique_id] = InvocationsSchema(
+                    fail_rate=round(totals.errors / len(test_invocations), 2) if test_invocations else 0,
+                    totals=totals,
+                    invocations=test_invocations,
+                    description=self._get_invocations_description(totals)
+                )
+            except Exception:
+                logger.error(f"Could not parse test ({invocations.get('test_unique_id')}) invocations - continue to the next test")
```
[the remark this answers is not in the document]
IDoneShaveIt: Will make it in another PR

PR #100, opened 2022-07-04
Slack alerts improvements

elongl:


monitor/test_result.py
```diff
@@ -32,22 +38,30 @@ def create_test_result_from_dict(test_result_dict: dict) -> Optional['TestResult
             else:
                 return ElementaryTestResult(**test_result_dict)
         except Exception:
-            logger.error(f"Failed parsing test result - {json.dumps(test_result_dict)}")
+            logger.exception(f"Failed parsing test result - {json.dumps(test_result_dict)}")
```
IDoneShaveIt, line 41: Usually its a good practice, but in this case the user will see the error exception itself and might get confused.
This exception is not breaking the flow, it will send (in the future) an "error" alert with the other alerts.
So printing the trace of the error when a user runs the edr can create a feeling that he shouldn't trust the edr.

monitor/test_result.py
```diff
@@ -32,22 +38,30 @@ def create_test_result_from_dict(test_result_dict: dict) -> Optional['TestResult
...
             return None
 
     def to_slack_message(self, slack_workflows: bool = False) -> dict:
-        pass
+        raise NotImplementedError
```
IDoneShaveIt, line 45: Like!

monitor/test_result.py
```diff
@@ -90,35 +105,38 @@ def description_display_name(description: str, default_description: str = '') ->
...
+            status
     ) -> None:
         super().__init__(id, model_unique_id, test_unique_id, status)
         self.test_unique_id = test_unique_id
         self.test_type = test_type
         self.database_name = database_name
         self.schema_name = schema_name
         self.table_name = table_name
-        self.table_full_name = '.'.join([database_name, schema_name, table_name]).lower()
+        table_full_name_parts = [database_name, schema_name]
```
IDoneShaveIt, line 135: NIT: I think it should be a method that is called in the init.
IMO init should be straight forward and simple as possible.
Adding if statements inside of it (that are not inline if statements) can make it a bit messy.

IDoneShaveIt (commented):
Nice work!
2 small comments 🙂

monitor/test_result.py
elongl: On one hand, I definitely agreed, that was my instinct as well.
On the other, I figured it's better to follow the pattern (there are already if statements within the constructor).
Either make all of them external functions, or keep all of the logic within the constructor, but I think it's even worse to mix the two. What do you think?

IDoneShaveIt: I think its better to start the change some where (even if not improving everything) instead of making the original "problem" bigger.
But, this is my opinion and this is a NIT, so I am fine with leaving it like this for now

IDoneShaveIt (approved):
LGTM

PR #101, opened 2022-07-05
Slack alerts stability

elongl:


monitor/data_monitoring.py
```diff
@@ -77,25 +88,49 @@ def _query_alerts(self, days_back: int) -> list:
...
+                    f"Could not sent the alert - {alert.id}. Full alert: {json.dumps(dict(alert_slack_message))}")
                 self.success = False
-        
+
+        sent_alert_count = len(sent_alerts)
+        self.execution_properties['sent_alert_count'] = sent_alert_count
+        if sent_alert_count > 0:
+            self._update_sent_alerts(sent_alerts)
+
+    def _send_raw_dicts_to_slack(self, alert_dicts: List[dict]) -> None:
```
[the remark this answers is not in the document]
IDoneShaveIt: So this method is really similar to the `_send_to_slack` method.
the only thing that actually different is the way we create the message itself.
I think it will be better to merge those methods into one and create a method for creating the message.
Plus in your case, the error alerts won't be calculated inside of the `alerts_with_progress_bar` which we do want to count :)

IDoneShaveIt (commented):
Nice PR 🙂 
I think we should align our typing conventions, and I left a comment about the error message sending method.

monitor/data_monitoring.py
```diff
@@ -64,43 +80,68 @@ def _update_sent_alerts(self, alert_ids) -> None:
...
+                test_result.generate_slack_message(is_slack_workflow=self.config.is_slack_workflow)
+            )
+            for test_result in test_results
+        ]
+
+    @staticmethod
+    def _get_slack_alert_from_dict(alert_dicts: List[Dict[str, Any]]):
+        return [
+            SlackAlert(
+                alert_dict['id'],
```
oravi, line 133: Use get here? (I know it should always be there but just in case as this is the last resort :))

monitor/data_monitoring.py
```diff
@@ -64,43 +80,69 @@ def _update_sent_alerts(self, alert_ids) -> None:
...
                 channel_name=self.slack_channel_name,
-                message=alert_slack_message
+                message=alert.message
             )
             if sent_successfully:
                 sent_alerts.append(alert.id)
             else:
-                logger.error(f"Could not sent the alert - {alert.id}. Full alert: {json.dumps(dict(alert_slack_message))}")
+                logger.error(
+                    f"Could not sent the alert - {alert.id}. Full alert: {json.dumps(dict(alert.message))}")
```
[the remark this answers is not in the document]
elongl: Wasn't mine 😉

monitor/data_monitoring.py
elongl: I can't, the entire logic would break if an alert doesn't have an alert ID.  
How would we flag that it was sent?

monitor/test_result.py
```diff
@@ -138,17 +138,20 @@ def __init__(
...
-            self.detected_at_utc = detected_at_utc.strftime('%Y-%m-%d %H:%M:%S')
-            self.detected_at = convert_utc_time_to_local_time(detected_at_utc).strftime('%Y-%m-%d %H:%M:%S')
-        self.owners = try_load_json(owners) if owners else ''
+            try:
+                detected_at_utc = datetime.fromisoformat(detected_at)
+                self.detected_at_utc = detected_at_utc.strftime('%Y-%m-%d %H:%M:%S')
+                self.detected_at = convert_utc_time_to_local_time(detected_at_utc).strftime('%Y-%m-%d %H:%M:%S')
+            except (ValueError, TypeError):
+                logger.error(f'Failed to parse "detect_at" field.')
+        self.owners = try_load_json(owners) or ''
```
oravi, line 147: I think that maybe we should do this owners and tags parsing inside 'to_slack_message' so it won't impact the report? I think the repot expects a list and here in the alerts we convert it to string

monitor/test_result.py
```diff
@@ -138,17 +138,20 @@ def __init__(
         self.table_full_name = '.'.join(table_full_name_parts).lower()
         self.detected_at = None
         if detected_at:
```
oravi, line 140: I know it has been this way even before but maybe we should also init self.detected_at_utc with None before this if statement?

PR #103, opened 2022-07-06
Users are recommended to upgrade version when available.

elongl:


elongl:
image

cli/cli.py
```diff
@@ -46,6 +49,16 @@ def format_help(self, ctx, formatter):
         self.format_options(ctx, formatter)
         self.format_epilog(ctx, formatter)
 
+    def recommend_version_upgrade(self):
+        latest_version = package.get_latest_package_version()
+        current_version = package.get_package_version()
+        if current_version != latest_version:
```
IDoneShaveIt, line 55: I think that making it smaller is better :)
While developing our version could be higher

utils/package.py
```diff
@@ -9,4 +10,13 @@ def get_package_version() -> Optional[str]:
...
\ No newline at end of file
+    return None
+
+
+def get_latest_package_version() -> Optional[str]:
+    resp = requests.get('https://pypi.org/pypi/elementary-data/json')
+    try:
+        resp.raise_for_status()
+        return resp.json()['info']['version']
+    except requests.HTTPError:
```
IDoneShaveIt, line 21: Lets make the exception general :)
It could be that in the future pypi will change the response structure and we will get a different error that will break the cli

utils/package.py
```diff
@@ -9,4 +10,13 @@ def get_package_version() -> Optional[str]:
...
-    return None
\ No newline at end of file
+    return None
+
+
+def get_latest_package_version() -> Optional[str]:
+    resp = requests.get('https://pypi.org/pypi/elementary-data/json')
+    try:
+        resp.raise_for_status()
+        return resp.json()['info']['version']
```
IDoneShaveIt, line 20: .get is always safer :)

IDoneShaveIt (commented):
I like this PR!
Small comments :)

IDoneShaveIt (approved):
LGTM

PR #104, opened 2022-07-06
Added a script for extracting the Elementary alerts from a dbt.log.

elongl:


scripts/dbt_log_to_elementary_alerts.py
```diff
@@ -0,0 +1,37 @@
+import json
+import sys
+
+
+def get_elementary_log_lines(dbt_log: str):
+    return [line for line in dbt_log.splitlines() if 'Elementary' in line]
```
oravi, line 6: Maybe use 'Elementary:' (without the ':' you could catch other lines with Elementary inside that are not intended to be parsed as macro result)

PR #105, opened 2022-07-06
Ensuring that a Slack connection is provided when needed.

elongl:


monitor/cli.py
```diff
@@ -1,19 +1,24 @@
...
-from config.config import Config
-from monitor.data_monitoring import DataMonitoring
+from utils.package import get_package_version
 
 yaml = OrderedYaml()
 
+logger = get_logger(__name__)
 
-def get_cli_properties() -> dict:
+ERR_MISSING_SLACK_CONN = 'Either a Slack token or webhook is required to send the report.'
```
oravi, line 18: cosmetics - since you use this message in the 'monitor' command maybe say something - 
'Either a Slack token or webhook is required'

PR #112, opened 2022-07-07
Fix a unit test due to the wrong dict key

yu-iskw:
SSIA


IDoneShaveIt (approved):
Thanks for fixing it!
Merging 🙂

PR #113, opened 2022-07-07
Add a github action to run unit tests

yu-iskw:
## Motivation
As far as I know, we don't have any CI jobs to run unit tests. I would like to introduce a CI job using GitHub Actions so that we continuously keep the source code quality great.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -1,0 +1,27 @@
+name: Unit tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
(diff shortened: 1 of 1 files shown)


yu-iskw:
@Maayan-s What do you think of the idea? The action might not be perfect, but it would be worthwhile introducing continuous integration.

IDoneShaveIt:
Hey @yu-iskw!
Thanks for another contribution!
This is a really good idea, and something we were talking about implementing.
Will go over your PR soon, and keep you updated!

PR #115, opened 2022-07-10
fixed vars params for cli monitor

IDoneShaveIt:
Fixed a bug in dbt-vars param at cli monitor:
- wrong use of yaml: `ordered_yaml -> yaml`
- add vars support to data_monitoring run method

### DIFF
--- a/monitor/data_monitoring.py
+++ b/monitor/data_monitoring.py
@@ -169,10 +169,9 @@ class DataMonitoring(object):
         if alerts:
             self._send_alerts_to_slack(alerts)
 
-    def run(self, days_back: int, dbt_full_refresh: bool = False, vars: Optional[dict] = None) -> bool:
-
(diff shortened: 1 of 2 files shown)


monitor/data_monitoring.py
```diff
@@ -169,7 +169,7 @@ def _send_alerts(self, days_back: int):
         if alerts:
             self._send_alerts_to_slack(alerts)
 
-    def run(self, days_back: int, dbt_full_refresh: bool = False) -> bool:
+    def run(self, days_back: int, dbt_full_refresh: bool = False, vars: Optional[dict] = None) -> bool:
```
elongl, line 172: what do you think about renaming `vars` to `dbt_vars` to avoid the shadowed name with the global `vars`, and also to remain consistent with `dbt_full_refresh`?

IDoneShaveIt: agree

PR #116, opened 2022-07-10
Added support for non-test (models) alerts.

elongl:


monitor/alert.py
```diff
@@ -0,0 +1,459 @@
...
+        return ' '.join([word[0].upper() + word[1:] for word in str_value.split('_')])
+
+    @staticmethod
+    def description_display_name(description: str, default_description: str = '') -> str:
+        if description:
+            return description[0].upper() + description[1:]
+        return default_description
+
+
+class DbtTestAlert(TestAlert):
```
IDoneShaveIt, line 111: Wrap with `@dataclass` for typing

monitor/data_monitoring.py
```diff
@@ -64,8 +51,16 @@ def __init__(
         # slack client is optional
         self.slack_client = SlackClient.create_slack_client(self.slack_token, self.slack_webhook)
         self._download_dbt_package_if_needed(force_update_dbt_package)
+        self.elementary_database_and_schema = self.get_elementary_database_and_schema()
```
IDoneShaveIt, line 54: This should not run on the intitial.
It relevant only for alerts, and there is no need to run this when creating the report.

monitor/data_monitoring.py
```diff
@@ -64,8 +51,16 @@ def __init__(
...
         self.success = True
 
+    def get_elementary_database_and_schema(self):
+        try:
+            database_and_schema = self.dbt_runner.run_operation('get_elementary_database_and_schema')[0]
```
IDoneShaveIt, line 59:
```suggestion
            database_and_schema = self.dbt_runner.run_operation(macro_name='get_elementary_database_and_schema')[0]
```
You can never know if run_operation will change, and the first argument will still be macro_name

monitor/data_monitoring.py
```diff
@@ -64,8 +51,16 @@ def __init__(
...
+            return '.'.join(json.loads(database_and_schema.replace("'", '"')))
+        except Exception:
+            return '<elementary_schema>'
```
IDoneShaveIt, line 62: I think it should be <data_base>.<elementary_schema>?

monitor/data_monitoring.py
```diff
@@ -76,74 +71,80 @@ def _split_list_to_chunks(items: list, chunk_size: int = 50) -> List[List]:
...
-
-    def _send_alerts_to_slack(self, alerts: List[SlackAlert]) -> None:
-        sent_alerts = []
+            self.dbt_runner.run_operation(
+                macro_name='update_sent_alerts',
+                macro_args={'alert_ids': alert_ids_chunk, 'table_name': table_name},
+                json_logs=False
+            )
+
+    def _query_alerts(self, days_back: int) -> Alerts:
```
IDoneShaveIt, line 83: WDYT about making it a bit more structured and organized?
After splitting the data monitoring into Alerts and Report, we can have a BaseAlerts class that has dbt_runner, _query_alerts method, _query_alert_type method.
Then we can create TestAlerts, ModelAlerts, etc that inheritance from the BaseAlerts method.
That way each class is responsible for its own type of alerts, and is separated from the rest of the alerts types (will help disabling/enabling alerts by type + creating new alerts types).
Then we will have a workflow that calls the relevant alerts.

IDoneShaveIt (commented):
Nice work!
The changes are overall good, but I think that we are at the point where we need to split some files + classes.
Lets discuss it 🙂

monitor/data_monitoring.py
```diff
@@ -210,13 +207,12 @@ def generate_report(self, days_back: Optional[int] = None, test_runs_amount: Opt
...
                     </script>
                 """
-            elementary_html_file_name = f"elementary - {now_utc} utc.html".replace(" ", "_").replace(":", "-")
-            elementary_html_path = os.path.join(self.config.target_dir, elementary_html_file_name)
-            with open(elementary_html_path, 'w') as elementary_output_html_file:
-                elementary_output_html_file.write(elementary_output_html)
-            with open(os.path.join(self.config.target_dir, 'elementary_output.json'), 'w') as \
-                    elementary_output_json_file:
-                elementary_output_json_file.write(elementary_output_str)
+        elementary_html_path = self._get_report_file_path(now_utc, file_path)
```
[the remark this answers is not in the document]
elongl: It will crash as it naturally should, would you like me to wrap it in a more verbose exception?

monitor/data_monitoring.py
elongl: `run-operation` is a command that runs a macro.
I feel confident in using it as such and treating the first argument as the macro's name.

monitor/data_monitoring.py
elongl: Totally right!

monitor/alert.py
```diff
@@ -0,0 +1,452 @@
+import json
```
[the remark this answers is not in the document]
elongl: This is a new file 😮 
By splitting it you mean a file for each alert type? (`TestAlert`, `ModelAlert`, etc.)

monitor/data_monitoring.py
```diff
@@ -27,19 +26,7 @@
...
-
-
-@dataclass
-class SlackAlert:
-    id: str
-    message: SlackMessageSchema
-
-
-class DataMonitoring(object):
+class DataMonitoring:
```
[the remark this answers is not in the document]
elongl: Definitely agree, I'll attempt to make some changes regarding decoupling the alerts and report from the data monitoring.

monitor/data_monitoring.py
elongl: Interesting. Could you please elaborate on the difference between what you're talking about and the current alerts model? As far as I see, `Alert` is your `BaseAlert` and there's `TestAlert` and `ModelAlert` that inherit from it.

monitor/alert.py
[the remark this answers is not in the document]
IDoneShaveIt: Yeah :)

monitor/data_monitoring.py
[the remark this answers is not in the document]
IDoneShaveIt: I don't mind it crashing the run, I just want it to happen before we generate the report which could take some time (which can make it really frustrating to wait all that time just to realize you provided a broken path).

We do need to validate/crash the run before we generate the report itself + create a more verbose exception

monitor/data_monitoring.py
IDoneShaveIt: So currently Alert is more of a schema/class that represents a single alert data with formatting options, while BaseAlerts is a base class for handling multiple alerts of the same type like Test, Model, etc.
It shares the base logic for handling alerts like querying it, parsing it, and handling broken alerts that couldn't be formatted.

monitor/data_monitoring.py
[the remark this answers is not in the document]
elongl: Agreed! Updated.

PR #120, opened 2022-07-13
hotfix/add backfill to alerts model

IDoneShaveIt:
add backfill to alerts model to prevent race condition when 2 dbt test runs + edr monitor

### DIFF
--- a/monitor/dbt_project/models/alerts.sql
+++ b/monitor/dbt_project/models/alerts.sql
@@ -31,7 +31,7 @@ from all_alerts
 {%- if is_incremental() %}
 {%- set row_count = elementary.get_row_count(this) %}
     {%- if row_count > 0 %}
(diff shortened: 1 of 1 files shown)


monitor/dbt_project/models/alerts.sql
```diff
@@ -31,6 +31,7 @@ from all_alerts
 {%- if is_incremental() %}
 {%- set row_count = elementary.get_row_count(this) %}
     {%- if row_count > 0 %}
-        where detected_at > (select max(detected_at) from {{ this }})
+        {# We use a backfill of 2 days to prevent race condition when 2 dbt tests + edr monitor runs together #}
```
elongl, line 34:
```suggestion
        {# We use a backfill of 2 days to prevent race condition when multiple dbt tests + edr monitor run together #}
```

PR #125, opened 2022-07-14
Add node types to report lineage

IDoneShaveIt:
- add node types to report lineage 
- fix model depends on nodes query to include sources and exposures as well


elongl:
LGTM!

PR #135, opened 2022-07-20
Add exposures to report and other improvements 

IDoneShaveIt:
- Added exposures to the models object in the report
- Removed seeds from the report lineage
- Fixed send-report bug (implemented support for pagination in slack client)
- Fixed CLI bug where we throw error when slack token/webhook is provided in the config

### DIFF
--- a/clients/slack/slack_client.py
+++ b/clients/slack/slack_client.py
@@ -110,6 +110,7 @@ class SlackWebClient(SlackClient):
             return channels
         except SlackApiError as e:
             logger.error(f"Elementary app failed to query all Slack public channels. Error: {e}")
+            return []
 
 class SlackWebhookClient(SlackClient):
     def _initial_client(self):
--- a/clients/slack/slack_client.py
+++ b/clients/slack/slack_client.py
@@ -78,15 +78,16 @@ class SlackWebClient(SlackClient):
 
     def _get_channel_id(self, channel_name: str) -> str:
         try:
+            channels = self._get_channels()
             channel = [
                 channel
                 for channel
-                in self.client.conversations_list()["channels"]
+                in channels
                 if channel["name"] == channel_name
             ][0]
             return channel["id"]
-        except Exception as e:
-            logger.error(f"Could not find the channel - {channel_name}. Error: {e}")
+        except Exception:
+            logger.error(f"Could not find the channel - {channel_name}. Elementary app's available cahnnels: {[channel.get('name') for channel in channels]}")
 
     def _join_channel(self, channel_id: str) -> bool:
         try:
@@ -96,6 +97,19 @@ class SlackWebClient(SlackClient):
             logger.error(f"Elementary app failed to join the given channel. Error: {e}")
             return False
 
+    def _get_channels(self) -> list[dict]:
+        try:
+            channels = []
+            has_more = True
+            cursor = None
+            while has_more:
+                response = self.client.conversations_list(cursor=cursor)
+                channels.extend(response["channels"])
+                cursor = response.get("response_metadata", {}).get("next_cursor")
+                has_more = True if cursor else False
+            return channels
+        except SlackApiError as e:
+            logger.error(f"Elementary app failed to query all Slack public channels. Error: {e}")
 
 class SlackWebhookClient(SlackClient):
     def _initial_client(self):
--- a/monitor/api/lineage/schema.py
+++ b/monitor/api/lineage/schema.py
@@ -4,6 +4,8 @@ from pydantic.typing import Literal
 import networkx as nx
 import re
 
+SEED_PATH_PATERN = re.compile(r"^seed\.")
+
 NodeUniqueIdType = str
 NodeType = Literal["model", "source", "exposure"] 
 
@@ -23,8 +25,7 @@ class NodeDependsOnNodesSchema(BaseModel):
     def _format_node_id(cls, node_id: str):
         # Currently we don't save seeds in our artifacts.
         # We remove seeds from the lineage graph (as long as we don't support them).
-        seed_patern = re.compile(r"^seed\.")
-        if re.search(seed_patern, node_id):
+        if re.search(SEED_PATH_PATERN, node_id):
             return None
         return node_id
 
--- a/monitor/api/lineage/schema.py
+++ b/monitor/api/lineage/schema.py
@@ -16,19 +16,16 @@ class NodeDependsOnNodesSchema(BaseModel):
     @validator("depends_on_nodes", pre=True, always=True)
     def set_depends_on_nodes(cls, depends_on_nodes):
         formatted_depends_on = depends_on_nodes or []
-        return [cls._format_node_id(node_id) for node_id in formatted_depends_on]
-    
-    @validator("unique_id", pre=True, always=True)
-    def set_unique_id(cls, unique_id):
-        return cls._format_node_id(unique_id)
-    
+        formatted_depends_on = [cls._format_node_id(node_id) for node_id in formatted_depends_on]
+        return [node_id for node_id in formatted_depends_on if node_id]
+
     @classmethod
     def _format_node_id(cls, node_id: str):
-        # Currently seeds are saved as sources.
-        # Convert seed prefix into source for nodes unique ids.
+        # Currently we don't save seeds in our artifacts.
+        # We remove seeds from the lineage graph (as long as we don't support them).
         seed_patern = re.compile(r"^seed\.")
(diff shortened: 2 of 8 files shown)


clients/slack/slack_client.py
```diff
@@ -78,15 +78,16 @@ def upload_file(self, channel_name: str, file_path: str, message: SlackMessageSc
 
     def _get_channel_id(self, channel_name: str) -> str:
         try:
+            channels = self._get_channels()
             channel = [
```
oravi, line 82: isn't it clearer to just iterate over the channels and return the first with a matching name?

clients/slack/slack_client.py
```diff
@@ -78,15 +78,16 @@ def upload_file(self, channel_name: str, file_path: str, message: SlackMessageSc
...
                 for channel
-                in self.client.conversations_list()["channels"]
+                in channels
                 if channel["name"] == channel_name
             ][0]
             return channel["id"]
-        except Exception as e:
-            logger.error(f"Could not find the channel - {channel_name}. Error: {e}")
+        except Exception:
+            logger.error(f"Could not find the channel - {channel_name}. Elementary app's available cahnnels: {[channel.get('name') for channel in channels]}")
```
oravi, line 90: what happens if the exception is in the get_channels?

monitor/api/lineage/schema.py
```diff
@@ -16,19 +16,16 @@ class NodeDependsOnNodesSchema(BaseModel):
...
+        formatted_depends_on = [cls._format_node_id(node_id) for node_id in formatted_depends_on]
+        return [node_id for node_id in formatted_depends_on if node_id]
+
     @classmethod
     def _format_node_id(cls, node_id: str):
-        # Currently seeds are saved as sources.
-        # Convert seed prefix into source for nodes unique ids.
+        # Currently we don't save seeds in our artifacts.
+        # We remove seeds from the lineage graph (as long as we don't support them).
         seed_patern = re.compile(r"^seed\.")
```
oravi, line 26: Maybe just check if the if the node_id starts with seed? because that's what this regex essentially validates. If you still want to keep the regex, move the compile outside of the function to be defined at the top of the class instead of in every invocation

monitor/api/models/models.py
```diff
@@ -45,51 +56,63 @@ def get_test_coverages(self) -> Dict[str, ModelCoverageSchema]:
         return coverages
 
     @staticmethod
-    def _normalize_dbt_model_dict(model: ModelSchema, is_source: bool = False) -> NormalizedModelSchema:
-        model_name = model.name
+    def _normalize_dbt_node_dict(
```
oravi, line 59: cosmetics - I'd rename to _normalize_dbt_artifact_dict

clients/slack/slack_client.py
IDoneShaveIt: Well yes, but I wanted to make it more generic.
Searching by id means we will never return the full list of channels (which is fine), and I thought it will be nice to implement it in a way other could use to :)

monitor/api/lineage/schema.py
IDoneShaveIt: Good point.
I like using regex, it is clear and super stricted :)
Will move the compile up!

monitor/api/models/models.py
```diff
@@ -45,51 +56,63 @@ def get_test_coverages(self) -> Dict[str, ModelCoverageSchema]:
...
-        normalized_model['tags'] = tags
-        normalized_model['model_name'] = model_name
-        normalized_model['normalized_full_path'] = ModelsAPI._normalize_model_path(
-            model_path=model.full_path,
-            model_package_name=model.package_name,
-            is_source=is_source
+        normalized_node = json.loads(node.json())
+        normalized_node['owners'] = owners
+        normalized_node['tags'] = tags
+        normalized_node['model_name'] = node_name
```
oravi, line 84: model_name is not relevant for exposures right?

monitor/api/models/schema.py
```diff
@@ -2,20 +2,30 @@
...
 
 
-class NormalizedModelSchema(ModelSchema):
+class ModelSchema(NodeSchema):
+    database_name: str
+    schema_name: str
+    table_name: str
+
+
+class ExposureSchema(NodeSchema):
```
oravi, line 21: all of the attributes below are mandatory? I think most of them are optional no?

monitor/api/models/schema.py
```diff
@@ -2,20 +2,30 @@
...
-class ModelSchema(BaseModel):
+class NodeSchema(BaseModel):
     name: str
     unique_id: str
-    database_name: str
-    schema_name: str
-    table_name: str
     owners: Optional[str]
     tags: Optional[str]
     package_name: str
```
oravi, line 10: I think package name is also optional

monitor/api/models/models.py
IDoneShaveIt: So everything got a name (exposures as well), the problem is that changing it into node/artifact name will break the UI.
So for now I wanted to leave it like that.
Will add a comment, and will talk with Nimrod.

monitor/api/models/schema.py
IDoneShaveIt: I compares it with dbt documentation.
What ever is optional I put as optional, everything else should be mandatory:)

monitor/api/models/schema.py
IDoneShaveIt: I am not sure, but will change for just in case

monitor/api/models/schema.py
IDoneShaveIt: Tested the fields and updated

monitor/api/models/models.py
```diff
@@ -45,51 +56,65 @@ def get_test_coverages(self) -> Dict[str, ModelCoverageSchema]:
...
+            artifact_package_name=artifact.package_name,
+            type=type
         )
-        return NormalizedModelSchema(**normalized_model)
+
+        if type == "exposure":
+            return NormalizedExposureSchema(**normalized_artifact)
+        elif type == "model":
+            return NormalizedModelSchema(**normalized_artifact)
+        elif type == "source":
```
oravi, line 95: What happens if the type is not one of them? what do you return in that case? also, I think Elon adds a new type of snapshot to dbt_models, let's verify with him?

monitor/api/models/schema.py
```diff
@@ -1,23 +1,41 @@
...
+
+class SourceSchema(ArtifactSchema):
+    database_name: str
+    schema_name: str
+    table_name: str
+
+
+class ExposureSchema(ArtifactSchema):
+    url: Optional[str]
+    type: Literal["dashboard", "notebook", "analysis", "ml", "application"]
```
oravi, line 29: Are you sure type and owner_email are not optional?

monitor/dbt_project/macros/get_exposures.sql
```diff
@@ -0,0 +1,28 @@
...
+                  tags,
+                  package_name,
+                  description,
+                  original_path as full_path
+                from {{ dbt_exposures_relation }}
+              )
+
+             select * from dbt_artifacts_exposures
+        {% endset %}
+        {% set exposures_agate = run_query(get_exposures_query) %}
```
oravi, line 24: what happens if dbt_exposures is empty?

clients/slack/slack_client.py
oravi: you can still get the channels from get_channels but here instead of iterating in a list comprehension you can just have a simple iteration

clients/slack/slack_client.py
oravi: But still if exception that is not SlackApi is thrown from get_channels it will be caught here and channels could be undefined, maybe just initialize it here with empty list before assigning a value to it from get_channels?

monitor/api/models/schema.py
IDoneShaveIt: 100%

monitor/api/models/models.py
IDoneShaveIt: Yeah but I can't change my PR before he merges his then.
Plus do we show snapshot as models in the sidebar/tables?
If not what is the case here?
I think that those question should be handled in Elon's PR.
If we change the models table in a PR, we need to change everything that is relied on it in that PR (models query, etc)

monitor/dbt_project/macros/get_exposures.sql
IDoneShaveIt: Nothing, it will return empty results and nothing will happen.
The report is working and everything is fine

PR #137, opened 2022-07-20
Non test alerts

elongl:


monitor/dbt_project/models/alerts/alerts_models.sql
```diff
@@ -2,17 +2,16 @@
...
-with alerts_dbt_models as (
-    select * from {{ ref('elementary', 'alerts_dbt_models') }}
-)
-
-select *, false as alert_sent from alerts_dbt_models
-
-{%- if is_incremental() %}
-    {{ get_new_alerts_where_clause(this) }}
-{%- endif %}
+{% set error_models_relation = adapter.get_relation(this.database, this.schema, 'error_models') %}
```
oravi, line 11: I'd check also if relation_exists and also if not, you need to return a select statement with the same schema so the macro that reads it will work (see empty_table.sql as a reference)

monitor/dbt_project/models/alerts/alerts.sql
```diff
@@ -0,0 +1,15 @@
...
+  config(
+    materialized = 'incremental',
+    unique_key = 'alert_id',
+    merge_update_columns = ['alert_id'],
+    on_schema_change = 'append_new_columns'
+  )
+}}
+
+
+select *, false as alert_sent from {{ ref('elementary', 'failed_tests') }}
```
oravi, line 11: This won't work with old package versions. I'd avoid adding this middle view for now and implementing its content here

monitor/dbt_project/models/alerts/alerts.sql
```diff
@@ -0,0 +1,15 @@
...
+
+{%- if is_incremental() %}
+    where {{ get_new_alerts_where_clause(this) }}
```
oravi, line 14: This macro introduced an issue if the row_count of 'this' is 0 - it returns none

monitor/dbt_project/macros/get_new_alerts_where_clause.sql
```diff
@@ -10,7 +10,7 @@
     {%- set row_count = elementary.get_row_count(alerts_model_relation) %}
```
oravi, line 10: I think instead of checking the row count we can check if last_alert_time_with_backfill is not none (we need to check what the macro result_value returns if it gets empty results). This will save a redundant query

monitor/dbt_project/macros/get_new_alerts_where_clause.sql
```diff
@@ -10,7 +10,7 @@
...
     {%- if row_count > 0 %}
         {% set where_clause %}
-            where detected_at > {{ last_alert_time_with_backfill }}
+            detected_at > {{ last_alert_time_with_backfill }}
```
oravi, line 13: also instead of the inner subquery - why don't we just check if detected_at > {{ last_alert_time_with_backfill }} and alert_sent = false?

monitor/alerts/model.py
```diff
@@ -30,22 +31,41 @@ def __post_init__(self):
         self.tags = prettify_json_str_set(self.tags)
 
     def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
+        if is_slack_workflow:
+            return SlackMessageSchema(text=json.dumps(self.__dict__))
+        if self.materialization == 'snapshot':
+            return self._snapshot_to_slack()
+
```
oravi, line 38: cosmetics - maybe add the code below to _model_to_slack

monitor/dbt_project/models/alerts/alerts_models.sql
elongl: It [seems](https://github.com/dbt-labs/dbt-core/blob/8442fb66a591c4f5353f8baaaaf62f846c5f5171/core/dbt/include/global_project/macros/adapters/relation.sql#L70) that `load_relation()` is just a user-friendly way of invoking `get_relation()`.  
Could you please elaborate a bit on why do it?

monitor/dbt_project/macros/get_new_alerts_where_clause.sql
oravi: Got it

elongl: @IDoneShaveIt Is there a good reason to have the `alert_sent = true` here?

oravi (approved):

monitor/dbt_project/macros/get_new_alerts_where_clause.sql
IDoneShaveIt: Yeah, because if you won't filter those alerts that already sent, the merging will override the existing sent ones, and will create them with false.
This will cause us sending the same alerts multiple times.

Merging in incremental delete the exist row, and write the new one (on redshift when they don't really support merge).

elongl:
1. Move subscribers to appear next to Owners.
2. Add subscribers to Snapshots.

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -1,19 +1,24 @@
...
+            where alert_sent = false and detected_at >= {{ get_alerts_time_limit(days_back) }}
         ),
-
-        models AS (
-            SELECT * FROM {{ ref('elementary', 'dbt_models') }}
+        models as (
+            select * from {{ ref('elementary', 'dbt_models') }}
+        ),
+        snapshots as (
+            select * from {{ ref('elementary', 'dbt_snapshots') }}
```
oravi, line 12: dbt_snapshot is a new model, so accessing it here means that the CLI won't work with old dbt package versions. I think that you need to first check if the table exists using the get_relation

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -1,19 +1,24 @@
...
         )
 
-        SELECT
+        select
             alerts.*,
-            models.meta as meta
-        FROM alerts LEFT JOIN models on (alerts.unique_id = models.unique_id)
+            models.meta as model_meta,
+            snapshots.meta as snapshot_meta
```
oravi, line 18: I think that we don't need two separate fields here because if you can't have both model_meta and shanpshot_meta. I'm suggesting having one meta field that is either the snapshots.meta or the models.meta

monitor/api/alerts.py
```diff
@@ -59,35 +59,42 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
...
-            model_meta = try_load_json(normalized_alert.get('model_meta'))
-            model_meta = model_meta if model_meta else {}
+            test_meta = try_load_json(normalized_alert.get('test_meta')) or {}
+            model_meta = try_load_json(normalized_alert.get('model_meta')) or {}
+            snapshot_meta = try_load_json(normalized_alert.get('snapshot_meta')) or {}
 
             subscribers = []
-            direct_subscribers = test_meta.get('subscribers', [])
+            test_subscribers = test_meta.get('subscribers', [])
             model_subscribers = model_meta.get('subscribers', [])
```
oravi, line 73: I think we should have only meta and test_meta instead of having model_meta and snapshot_meta and test_meta as you won't have model_meta and snapshot_meta together, it's either you have model_meta or snapshot_meta and therefore there should be only 'meta' that contains either one of them (also explained below).

monitor/dbt_project/macros/get_new_model_alerts.sql
elongl: How would you do it at the SQL-level?
Though, I can change:
```py
                                 'model_meta': elementary.insensitive_get_dict_value(model_result_alert_dict, 'model_meta'),
                                 'snapshot_meta': elementary.insensitive_get_dict_value(model_result_alert_dict, 'snapshot_meta'),
```
To:
```
'meta': elementary.insensitive_get_dict_value(model_result_alert_dict, 'model_meta') or elementary.insensitive_get_dict_value(model_result_alert_dict, 'snapshot_meta'),
```

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -1,20 +1,36 @@
 {% macro get_new_model_alerts(days_back) %}
     -- depends_on: {{ ref('alerts_models') }}
+    {% set elementary_database, elementary_schema = elementary.get_package_database_and_schema() %}
```
oravi, line 3: I'd use elementary.target_database() and target.schema

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -1,20 +1,36 @@
...
+        models as (
+            select * from {{ ref('elementary', 'dbt_models') }}
         )
-
-        SELECT
-            alerts.*,
-            models.meta as meta
-        FROM alerts LEFT JOIN models on (alerts.unique_id = models.unique_id)
+        {% if snapshots_relation %}
+            ,snapshots as (
```
oravi, line 16: I'd do here a union instead of using coalesce

monitor/api/alerts.py
```diff
@@ -59,23 +59,22 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
...
     @classmethod
     def _normalize_alert(cls, alert: dict) -> dict:
         try:
             normalized_alert = copy.deepcopy(alert)
-            test_meta = try_load_json(normalized_alert.get('test_meta'))
-            test_meta = test_meta if test_meta else {}
-            model_meta = try_load_json(normalized_alert.get('model_meta'))
-            model_meta = model_meta if model_meta else {}
+            test_meta = try_load_json(normalized_alert.get('test_meta')) or {}
+            model_meta = try_load_json(normalized_alert.get('model_meta')) or {}
```
oravi, line 68: cosmetics - maybe just call it meta and add a comment that it could be either a model meta or a snapshot meta

monitor/dbt_project/macros/get_new_test_alerts.sql
```diff
@@ -1,53 +1,37 @@
 {% macro get_new_test_alerts(days_back, results_sample_limit = 5) %}
-    -- depends_on: {{ ref('alerts_tests') }}
+    -- depends_on: {{ ref('alerts') }}
     {% set select_new_alerts_query %}
-        WITH alerts AS (
-            SELECT * FROM {{ ref('alerts_tests') }}
-            WHERE alert_sent = FALSE and detected_at >= {{ get_alerts_time_limit(days_back) }}
+        with new_alerts as (
```
oravi, line 4: cosmetics - let's rename the cte to new_tests?

monitor/api/alerts.py
elongl: I wanted to make a clear distinction between the model's meta and the test's.

monitor/dbt_project/macros/get_new_test_alerts.sql
elongl: In the context of `get_new_test_alert()`, the name makes sense, don't you think?

monitor/api/alerts.py
oravi: ok by me :)

oravi (approved):

PR #138, opened 2022-07-21
Support subscribers and channels

IDoneShaveIt:
- Support subscribers for alerts (uses subscribers field inside meta field of the artifacts)
 - Support custom channel for specific alerts (uses subscribers field inside meta field of the artifacts)
 
image

### DIFF
--- a/monitor/alerts/alert.py
+++ b/monitor/alerts/alert.py
@@ -14,7 +14,8 @@ class Alert:
         id: str,
         elementary_database_and_schema: str,
         subscribers: Optional[List[str]] = None,
-        slack_channel: Optional[str] = None
+        slack_channel: Optional[str] = None,
+        **kwargs
     ) -> None:
         self.id = id
         self.elementary_database_and_schema = elementary_database_and_schema
--- a/monitor/alerts/alert.py
+++ b/monitor/alerts/alert.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from typing import List, Optional
 
 from slack_sdk.models.blocks import SectionBlock
 
@@ -8,10 +8,18 @@ from utils.log import get_logger
 logger = get_logger(__name__)
 
 
-@dataclass
 class Alert:
-    id: str
-    elementary_database_and_schema: str
+    def __init__(
+        self,
+        id: str,
+        elementary_database_and_schema: str,
+        subscribers: Optional[List[str]] = None,
+        slack_channel: Optional[str] = None
+    ) -> None:
+        self.id = id
+        self.elementary_database_and_schema = elementary_database_and_schema
+        self.subscribers = subscribers
+        self.slack_channel = slack_channel
 
     _LONGEST_MARKDOWN_SUFFIX_LEN = 3
     _CONTINUATION_SYMBOL = '...'
--- a/monitor/alerts/malformed.py
+++ b/monitor/alerts/malformed.py
@@ -12,7 +12,8 @@ class MalformedAlert(Alert):
         elementary_database_and_schema: str,
         data: dict,
         subscribers: Optional[List[str]] = None,
-        slack_channel: Optional[str] = None
+        slack_channel: Optional[str] = None,
+        **kwargs
     ) -> None:
         super().__init__(id, elementary_database_and_schema, subscribers, slack_channel)
         self.data = data
--- a/monitor/alerts/malformed.py
+++ b/monitor/alerts/malformed.py
@@ -1,13 +1,22 @@
 import json
-from dataclasses import dataclass
+from typing import List, Optional
 
 from clients.slack.schema import SlackMessageSchema
 from monitor.alerts.alert import Alert
 
 
-@dataclass
 class MalformedAlert(Alert):
-    data: dict
+    def __init__(
+        self,
+        id: str,
+        elementary_database_and_schema: str,
+        data: dict,
+        subscribers: Optional[List[str]] = None,
+        slack_channel: Optional[str] = None
+    ) -> None:
+        super().__init__(id, elementary_database_and_schema, subscribers, slack_channel)
+        self.data = data
+
 
     def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
         return SlackMessageSchema(
--- a/monitor/alerts/model.py
+++ b/monitor/alerts/model.py
@@ -25,7 +25,8 @@ class ModelAlert(Alert):
         tags: str,
         status: str,
         subscribers: Optional[List[str]] = None,
-        slack_channel: Optional[str] = None
+        slack_channel: Optional[str] = None,
+        **kwargs
     ) -> None:
         super().__init__(id, elementary_database_and_schema, subscribers, slack_channel)
         self.unique_id = unique_id
--- a/monitor/alerts/model.py
+++ b/monitor/alerts/model.py
@@ -1,7 +1,7 @@
 import json
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Union
+from typing import List, Union
 
 from clients.slack.schema import SlackMessageSchema
 from monitor.alerts.alert import Alert
@@ -22,12 +22,14 @@ class ModelAlert(Alert):
     owners: str
     tags: str
     status: str
+    subscribers: List[str]
 
     TABLE_NAME = 'alerts_models'
 
     def __post_init__(self):
         self.owners = prettify_json_str_set(self.owners)
         self.tags = prettify_json_str_set(self.tags)
+        self.subscribers = prettify_json_str_set(self.subscribers)
 
     def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
         icon = ':small_red_triangle:'
@@ -41,7 +43,7 @@ class ModelAlert(Alert):
                                               [f'*Model*\n{self.alias}',
                                                f'*When*\n{self.detected_at}'],
                                               divider=True)
-        self._add_fields_section_to_slack_msg(slack_message, [f'*Owners*\n{self.owners}', f'*Tags*\n{self.tags}'])
+        self._add_fields_section_to_slack_msg(slack_message, [f'*Owners*\n{self.owners}', f'*Subscribers*\n{self.subscribers}' f'*Tags*\n{self.tags}'])
         self._add_text_section_to_slack_msg(slack_message, f'*Error Message*\n```{self.message}```')
         self._add_fields_section_to_slack_msg(slack_message,
                                               [f'*Full Refresh*\n{self.full_refresh}', f'*Path*\n{self.path}'],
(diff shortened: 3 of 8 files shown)


monitor/alerts/alert.py
```diff
@@ -8,10 +8,18 @@
 logger = get_logger(__name__)
 
 
-@dataclass
 class Alert:
-    id: str
-    elementary_database_and_schema: str
+    def __init__(
```
IDoneShaveIt, line 12: Convert from dataclass into class.
1. IMO dataclass should be used in very specific situations when the class is proper dataclass (and usually it will be better to use pydantic schema)
2. dataclass and inheritance doesn't work so well, and we inheritance (with default values) this class.

monitor/alerts/malformed.py
```diff
@@ -1,13 +1,22 @@
...
+from typing import List, Optional
 
 from clients.slack.schema import SlackMessageSchema
 from monitor.alerts.alert import Alert
 
 
-@dataclass
 class MalformedAlert(Alert):
-    data: dict
+    def __init__(
```
IDoneShaveIt, line 9: Convert from dataclass into class.
1. IMO dataclass should be used in very specific situations when the class is proper dataclass (and usually it will be better to use pydantic schema)
2. dataclass and inheritance doesn't work so well, and we inheritance (with default values) this class.

monitor/alerts/model.py
```diff
@@ -1,33 +1,51 @@
...
-    materialization: str
-    detected_at: Union[str, datetime]
-    database_name: str
-    schema_name: str
-    message: str
-    full_refresh: bool
-    owners: str
-    tags: str
-    status: str
+    def __init__(
```
IDoneShaveIt, line 11: Convert from dataclass into class.
1. IMO dataclass should be used in very specific situations when the class is proper dataclass (and usually it will be better to use pydantic schema)
2. dataclass and inheritance doesn't work so well, and we inheritance (with default values) this class.

monitor/alerts/test.py
```diff
@@ -15,11 +14,21 @@
 logger = get_logger(__name__)
 
 
-@dataclass
 class TestAlert(Alert):
-    model_unique_id: str
-    test_unique_id: str
-    status: str
+    def __init__(
```
IDoneShaveIt, line 18: Convert from dataclass into class.
1. IMO dataclass should be used in very specific situations when the class is proper dataclass (and usually it will be better to use pydantic schema)
2. dataclass and inheritance doesn't work so well, and we inheritance (with default values) this class.

monitor/alerts/model.py
```diff
@@ -41,7 +59,10 @@ def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
                                               [f'*Model*\n{self.alias}',
                                                f'*When*\n{self.detected_at}'],
                                               divider=True)
-        self._add_fields_section_to_slack_msg(slack_message, [f'*Owners*\n{self.owners}', f'*Tags*\n{self.tags}'])
+        self._add_fields_section_to_slack_msg(slack_message,
+                                              [f'*Owners*\n{self.owners}', f'*Tags*\n{self.tags}'])
+        if self.subscribers:
+            self._add_fields_section_to_slack_msg(slack_message, [f'*Subscribers*\n{", ".join(set(self.subscribers))}'])
```
oravi, line 65: maybe put it together with column if you have it?

IDoneShaveIt: Not sure I understand the request 😅

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -9,6 +9,15 @@
     {% set new_alerts = [] %}
     {% for model_result_alert_dict in model_result_alert_dicts %}
         {% set status = elementary.insensitive_get_dict_value(model_result_alert_dict, 'status') | lower %}
+        {% set meta = fromjson(elementary.insensitive_get_dict_value(model_result_alert_dict, 'meta', '{}')) %}
```
oravi, line 12: I think meta will be a string of a dumped json (see insert_dicts). I think you need to parse it and we already do it in the python so maybe worth extracting the subscribers and slack channel there? you can also wrap it with a try except to avoid failing the alert because of a parsing issue

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -9,6 +9,15 @@
...
+        {% set slack_channel = elementary.insensitive_get_dict_value(meta, 'channel') %}
+        {% set direct_subscribers = elementary.insensitive_get_dict_value(meta, 'subscribers', []) %}
+        {% set subscribers = [] %}
+        {% if direct_subscribers is string %}
```
oravi, line 16: same here, isn't it safer to do in the python?

monitor/dbt_project/macros/get_new_model_alerts.sql
IDoneShaveIt: It will, this is why I use fromjson.
I am not sure why it is better to move the logic to the python outside of the macro.
Can you elaborate 🤔?

monitor/dbt_project/macros/get_new_test_alerts.sql
```diff
@@ -4,6 +4,33 @@
         SELECT * FROM {{ ref('alerts_tests') }}
         WHERE alert_sent = FALSE and detected_at >= {{ get_alerts_time_limit(days_back) }}
     {% endset %}
+    {% set select_nodes_meta_query %}
```
oravi, line 7: I think it makes more sense either to join with model_id and bring the meta or add it to the alert in the package - wdyt?

monitor/dbt_project/macros/get_new_model_alerts.sql
IDoneShaveIt: I am not sure, what scenario is worrying you?

monitor/dbt_project/macros/get_new_test_alerts.sql
IDoneShaveIt: All options are good IMO.
In the current situation where changes were about to occur in the package, I though that this is the most suitable solution.

monitor/api/alerts.py
```diff
@@ -40,21 +42,50 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
...
+                        slack_channel=notmalaized_alert.get('slack_channel')
                     ))
         if malformed_alerts:
             logger.error('Failed to parse some alerts.')
             self.success = False
         return AlertsQueryResult(alerts, malformed_alerts)
+    
+    @classmethod
+    def _normalize_alert(cls, alert: dict) -> dict:
+        normalized_alert = copy.deepcopy(alert)
```
oravi, line 65: why not doing it in place? deepcopy is slow

monitor/alerts/model.py
```diff
@@ -1,33 +1,52 @@
...
+        database_name: str,
+        schema_name: str,
+        message: str,
+        full_refresh: bool,
+        owners: str,
+        tags: str,
+        status: str,
+        subscribers: Optional[List[str]] = None,
+        slack_channel: Optional[str] = None,
+        **kwargs
```
oravi, line 29: Not sure why this kwargs is needed, let's talk?

monitor/alerts/model.py
oravi: Explained over slack, I suggested putting the subscribers and the test column if exists at the same row above the error message

monitor/dbt_project/macros/get_new_model_alerts.sql
```diff
@@ -1,8 +1,19 @@
...
+            WHERE alert_sent = FALSE and detected_at >= {{ get_alerts_time_limit(days_back) }}
+        ),
+
+        models AS (
+            SELECT * FROM {{ ref('elementary', 'dbt_models') }}
+        )
+
+        SELECT
+            alerts.*,
+            models.meta as meta
```
oravi, line 15: I think this is a bug because in the normalize_alert you extract the channel only from model_meta and not from meta

monitor/api/alerts.py
```diff
@@ -40,21 +42,50 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
         if raw_alerts:
             alert_dicts = json.loads(raw_alerts[0])
             for alert_dict in alert_dicts:
+                notmalaized_alert = self._normalize_alert(alert=alert_dict)
```
oravi, line 45: if normalize_alert throws an exception, we don't want to crash the entire run. IMO - we should keep it in the try-except and use the MalformedAlert on the original alert as a last resort. WDYT?

monitor/api/alerts.py
IDoneShaveIt: I hate mutables objects.
There are more chances for bugs and its just not worth it IMO.
I can change it if you want.
Deepcopy is slow, but we have small objects so I don't mind using it 🤷🏻‍♂️

monitor/alerts/model.py
IDoneShaveIt: Alerts are generated from one place.
So we have one method that passes all the data every alert type needs (this is one of the things we talked about).
So it is much safer to add this **kwargs.

monitor/alerts/model.py
IDoneShaveIt: It won't work well.
The column field is optional as well + might contain 2 message fields.
Doing this like that makes it messy (need to have if statements that looks at 3 different fields)
Ugly, and don't worth it at the time IMO

monitor/api/alerts.py
```diff
@@ -40,21 +42,53 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
...
+            else:
+                subscribers.append(model_subscribers)
+
+            slack_channel = model_meta.get('channel')
+
+            normalized_alert['subscribers'] = subscribers
+            normalized_alert['slack_channel'] = slack_channel
+            return normalized_alert
+        except Exception:
+            logger.error(f"Failed to normalaize the alert '{alert.get('id')}'. Might affect the rest of the flow of the alert")
```
oravi, line 90: Maybe say - "Failed to extract alert subscribers and alert custom slack channel {alert.get('id')}. Ignoring it for now and main slack channel will be used"

monitor/api/alerts.py
```diff
@@ -40,21 +42,53 @@ def _query_alert_type(self, run_operation_args: dict, alert_factory_func: Callab
...
         if malformed_alerts:
             logger.error('Failed to parse some alerts.')
             self.success = False
         return AlertsQueryResult(alerts, malformed_alerts)
+    
+    @classmethod
+    def _normalize_alert(cls, alert: dict) -> dict:
+        try:
+            normalized_alert = copy.deepcopy(alert)
+            meta = try_load_json(normalized_alert.get('meta'))
```
oravi, line 67: maybe let's be explicit and call it 'test_meta' in all the relevant places?

PR #139, opened 2022-07-21
Flag sample data for passed elementary tests

IDoneShaveIt:
Added a flag to disable sampling for elementary passed tests.
An empty result test will look like:
image

### DIFF
--- a/monitor/api/tests/tests.py
+++ b/monitor/api/tests/tests.py
@@ -27,14 +27,14 @@ class TestsAPI(APIClient):
             self,
             days_back: Optional[int] = 7,
             metrics_sample_limit: int = 5,
-            disable_elementary_passed_tests_samples: bool = False
+            disable_passed_test_metrics: bool = False
     ) -> Dict[TestUniqueIdType, Dict[str, Any]]:
         run_operation_response = self.dbt_runner.run_operation(
             macro_name='get_tests_sample_data',
             macro_args=dict(
                 days_back=days_back,
                 metrics_sample_limit=metrics_sample_limit,
-                disable_elementary_passed_tests_samples=disable_elementary_passed_tests_samples
+                disable_passed_test_metrics=disable_passed_test_metrics
             )
         )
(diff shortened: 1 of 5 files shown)


monitor/api/tests/tests.py
```diff
@@ -26,13 +26,15 @@ def get_tests_metadata(self, days_back: Optional[int] = 7) -> List[TestMetadataS
     def get_tests_sample_data(
             self,
             days_back: Optional[int] = 7,
-            metrics_sample_limit: int = 5
+            metrics_sample_limit: int = 5,
+            disable_elementary_passed_tests_samples: bool = False
```
oravi, line 30: cosmetics - rename to disable_passed_test_metrics

monitor/cli.py
```diff
@@ -193,15 +194,21 @@ def monitor(
     type=str,
     help="The file path where Elementary's report will be saved."
 )
+@click.option(
+    '--disable-elementary-passed-tests-samples',
+    type=bool,
+    default=False,
+    help="If true, elementary won't query sample data for elementary passed tests (can improve report run time)."
```
oravi, line 201: cosmetics - 'If set to true elementary report won't show data metrics for passed tests (this can improve report creation time)'

monitor/dbt_project/macros/get_tests_sample_data.sql
```diff
@@ -35,7 +35,11 @@
         {% set status = elementary.insensitive_get_dict_value(test, 'status') | lower %}
 
         {% set test_rows_sample = none %}
-        {%- if (test_type == 'dbt_test' and status in ['fail', 'warn']) or (test_type != 'dbt_test' and status != 'error') -%}
+        {% set elementary_tests_filtered_status = ['error'] %}
```
oravi, line 38: cosmetics - I'd be explicit here and instead of using blacklist I'd use a whitelist - 
elementary_tests_whitelist_status = ['fail', 'warn'] and if the flag is not enabled then you can add 'pass'

oravi (approved):

PR #141, opened 2022-07-25
Not using the newly added default argument of a macro.

elongl:
In the new dbt package, the default argument of `get_package_database_and_schema` is `elementary`.  
However, if a user updates the CLI before the dbt package, this change wouldn't apply.  
This would still work and wouldn't crash for the user, it would just return `<elementary_database>.<elementary_schema>` because of this:

```py
        try:
            return utils.dbt.get_elementary_database_and_schema(self.dbt_runner)
        except Exception:
            logger.error("Failed to parse Elementary's database and schema.")
            return '<elementary_database>.<elementary_schema>'
```

This is only relevant if a user's query is too long and a query for the query is returned.


IDoneShaveIt (approved):
LGTM

PR #143, opened 2022-07-26
Created an Elementary package namespace.

elongl:
The package is now under a single `elementary` namespace rather than the global scope.


IDoneShaveIt (approved):
LGTM

PR #144, opened 2022-07-26
User ID is initialized upon package installation rather than first run.

elongl:
The User ID is generated at build-time rather than the first run-time.
This is useful for tracking virtualized environments such as Docker more consistently.


elementary/monitor/cli.py
```diff
@@ -207,9 +205,10 @@ def report(ctx, days_back, config_dir, profiles_dir, update_dbt_package, profile
     track_cli_start(anonymous_tracking, 'monitor-report', get_cli_properties(), ctx.command.name)
     try:
         data_monitoring = DataMonitoring(config, update_dbt_package)
-        success = data_monitoring.generate_report(days_back=days_back, test_runs_amount=executions_limit,
-                                                  file_path=file_path,
+        success = data_monitoring.generate_report(anonymous_tracking.anonymous_user_id, days_back=days_back,
```
IDoneShaveIt, line 208: please be specific when passing an argument to a method:
```suggestion
        success = data_monitoring.generate_report(user_id=anonymous_tracking.anonymous_user_id, days_back=days_back,
```
This is a better practice IMO - it is safer + it makes it really easy to understand what each argument is used for.

elementary/monitor/cli.py
```diff
@@ -317,9 +316,9 @@ def send_report(
             slack_channel_name=slack_channel_name
         )
         command_succeeded = False
-        generated_report_successfully, elementary_html_path = data_monitoring.generate_report(days_back=days_back,
-                                                                                              test_runs_amount=executions_limit,
-                                                                                              disable_passed_test_metrics=disable_passed_test_metrics)
+        generated_report_successfully, elementary_html_path = data_monitoring.generate_report(
+            anonymous_tracking.anonymous_user_id, days_back=days_back, test_runs_amount=executions_limit,
```
IDoneShaveIt, line 320: same as previous comment

IDoneShaveIt (approved):
Approved with small comment

PR #162, opened 2022-08-03
Upload report to cloud

elongl:
Added a new command (`upload-report`) that uploads the report to to an S3 or a GCS bucket.


config/config.py
```diff
@@ -1,71 +1,58 @@
...
+
+        self.target_dir = config.get('target-path') or os.getcwd()
+
+        self.slack_webhook = slack_webhook or config.get(self._SLACK, {}).get('notification_webhook')
+        self.slack_token = slack_token or config.get(self._SLACK, {}).get('token')
+        self.slack_channel_name = slack_channel_name or config.get(self._SLACK, {}).get('channel_name')
+        self.is_slack_workflow = config.get(self._SLACK, {}).get('workflows', False)
+
+        self.aws_profile_name = aws_profile_name or config.get(self._AWS, {}).get('profile_name')
+        self.aws_access_key_id = aws_access_key_id
```
elongl, line 27: These files are not retrieved from the configuration.
They could either be supplied via the CLI arguments, or read from `~/.aws/credentials`.

monitor/data_monitoring.py
```diff
@@ -156,25 +154,67 @@ def generate_report(self, days_back: Optional[int] = None, test_runs_amount: Opt
...
         return self.success
 
+    def upload_report(self, html_path: str):
+        if self.s3_client:
+            self._upload_report_to_s3(html_path)
+        if self.gcs_client:
+            self._upload_report_to_gcs(html_path)
+        return self.success
+
+    def _upload_report_to_s3(self, html_path: str):
```
IDoneShaveIt, line 184: I think that s3 client should be a class with a method for upload and a method for updating buckets website.
data_monitoring shouldn't know so much about s3, and shouldn't be aware about its internals.

monitor/data_monitoring.py
```diff
@@ -156,25 +154,67 @@ def generate_report(self, days_back: Optional[int] = None, test_runs_amount: Opt
...
+                self.s3_client.put_bucket_website(
+                    Bucket=self.config.s3_bucket_name,
+                    WebsiteConfiguration={'IndexDocument': {'Suffix': report_filename}}
+                )
+                logger.info("Updated S3 bucket's website.")
+        except botocore.exceptions.ClientError:
+            logger.error('Failed to upload report to S3.')
+            self.success = False
+
+    def _upload_report_to_gcs(self, html_path: str):
```
IDoneShaveIt, line 202: I think that gcs client should be a class with a method for upload and a method for updating buckets website.
data_monitoring shouldn't know so much about gcs, and shouldn't be aware about its internals.

clients/gcs/client.py
```diff
@@ -0,0 +1,9 @@
+from google.cloud import storage
+
+from config.config import Config
+
+
+def get_gcs_client(config: Config):
```
IDoneShaveIt, line 6: See my comments on data_monitoring about gcs

clients/s3/client.py
```diff
@@ -0,0 +1,12 @@
+import boto3
+
+from config.config import Config
+
+
+def get_s3_client(config: Config):
```
IDoneShaveIt, line 6: See my comments on data_monitoring about s3

setup.py
```diff
@@ -31,6 +31,8 @@
         'requests<3.0.0',
         'beautifulsoup4',
         'posthog',
+        'boto3',
+        'google-cloud-storage',
```
IDoneShaveIt, line 35: mmm maybe it will be better to make them as extras requirements?
I think that installing boto3/gcs when there is no use of it can be annoying.
WDYT?

IDoneShaveIt (commented):
Overall I like it!
Few comments :)

setup.py
elongl: I don't want our package to be _too_ modular in the sense that our users would need to think too much about which features to install and how.

I looked into the packages' requirements and they look very lean and only relevant to their domain so the chances of conflicts are quite low.

`google-cloud-storage`
Requires: google-api-core, google-auth, google-cloud-core, google-resumable-media, requests

`boto3`
Requires: botocore, jmespath, s3transfer

elongl:
@IDoneShaveIt I've applied your great suggestions! Please feel free to re-review when you have the time.

monitor/data_monitoring.py
```diff
@@ -156,23 +151,41 @@ def generate_report(self, days_back: Optional[int] = None, test_runs_amount: Opt
...
+            webbrowser.open_new_tab('file://' + html_path)
         self.execution_properties['report_end'] = True
         self.execution_properties['success'] = self.success
         return self.success, html_path
 
-    def send_report(self, elementary_html_path: str) -> bool:
-        if os.path.exists(elementary_html_path):
+    def send_report(self, html_path: str) -> bool:
+        if self.slack_client:
+            self.execution_properties['send_to_slack'] = True
```
IDoneShaveIt, line 162: let's add at the end of the if statement `self.execution_properties['sent_to_slack_successfully'] = file_uploaded_successfully`, just like you did with uploading to s3 and gcs 🙂

monitor/cli.py
```diff
@@ -329,5 +303,136 @@ def send_report(
...
+    help="The path to the Google service account json file"
+)
+@click.option(
+    '--gcs-bucket-name',
+    type=str,
+    default=None,
+    help="The name of the GCS bucket to upload the report to"
+)
+@click.pass_context
+def upload_report(
```
IDoneShaveIt, line 395: maybe it will be a good thing to merge between send report and upload report?
uploading report to slack is much like uploading it to s3/gcs?

monitor/cli.py
```diff
@@ -247,27 +237,17 @@ def report(ctx, days_back, config_dir, profiles_dir, update_dbt_package, profile
...
-    type=str,
-    default=None,
-    help="A slack webhook URL for sending alerts to a specific channel (also could be configured once in config.yml)"
-)
 @click.option(
     '--slack-token', '-st',
     type=str,
     default=None,
     help="A slack token for sending alerts over slack (also could be configured once in config.yml)",
-    cls=RequiredIf,
```
IDoneShaveIt, line 261: Don't we want the require if 🤔 ?
It lets users to know right in the start that they must provide both

monitor/cli.py
```diff
@@ -247,27 +237,17 @@ def report(ctx, days_back, config_dir, profiles_dir, update_dbt_package, profile
...
-    required_if="slack_channel_name"
 )
 @click.option(
     '--slack-channel-name', '-ch',
     type=str,
     default=None,
     help="The slack channel which all alerts will be sent to (also could be configured once in config.yml)",
-    cls=RequiredIf,
```
IDoneShaveIt, line 269: Don't we want the require if 🤔 ?
It lets users to know right in the start that they must provide both

monitor/cli.py
```diff
@@ -329,5 +303,136 @@ def send_report(
...
+    default='default',
+    help="AWS profile name for AWS",
+)
+@click.option(
+    '--aws-access-key-id',
+    type=str,
+    default=None,
+    help="The access key id for AWS"
+)
+@click.option(
```
IDoneShaveIt, line 370: See my comment about the require if 🙂

monitor/data_monitoring.py
```diff
@@ -156,23 +151,41 @@ def generate_report(self, days_back: Optional[int] = None, test_runs_amount: Opt
...
+            logger.info('Sent report to Slack.')
             if not file_uploaded_successfully:
+                logger.error('Failed to send report to Slack.')
                 self.success = False
-        else:
-            logger.error('Could not send Elementary monitoring report because it does not exist')
-            self.success = False
+        return self.success
+
+    def upload_report(self, html_path: str):
```
IDoneShaveIt, line 174: Can we merge between the upload report and the send report methods?
They are quite the same with other clients and supported services 🙂  
(adding an `if self.slack_client`)

IDoneShaveIt (commented):
Few more comments 🙂

monitor/data_monitoring.py
elongl: There's not much point because we'd see the command is `send-report` and we have the `success` variable which is equivalent.

monitor/cli.py
elongl: No because it doesn't allow users to mix.
For instance, I wanted to have my token in the config file but I wanted to be able to choose the channel easily within the CLI rather than edit the config file.

monitor/cli.py
elongl: This was my intuition as well.
Thinking it over, we got into a relatively complex option / decision matrix regarding what to do relative to the arguments that the user supplied. For instance, if he provided both S3 and Slack, then send the report to S3 and a link to the Slack, etc. But that creates a lot of complication, both in terms of development and for the user.

However, if we dumb it down so that you basically just send off the report to whichever method specified (Slack, S3, GCS) - I don't see why not and I think it'd be for the better. I also think the difference is too small to justify a separation.

The only part is the `--update-bucket-website` flag which might make _a bit_ less sense.

What do you think? @oravi @Maayan-s

oravi: I don't think that it makes sense to implement both in send-report. My argument is that send-report is not just uploading a report to slack, it will contain more information that will be added down the road like the summary report we have talked about. Also, I think that commands with too many arguments are confusing

elongl: I don't see how adding a summary appended to the report changes it.
It's just as if I would've uploaded the object to S3 with [metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html).
Those are just different mediums of sending the report to.

Regarding the too many arguments:
1. You could also claim the same for `upload-report` if we'd add more platforms, it's inevitable.
2. Realistically, every organization will have their favored method and they'll use only that.
3. We provide the config file for that.

monitor/data_monitoring.py
IDoneShaveIt: If we'll join the methods it will 🙂

Plus adding this info to the execution_properties can't hurt, and will make it more agnostic to changes in the future (it is likely to forget adding this data due to a change in the way we call the method in the future + if send-report cli command will have more steps in the future that take place after sending the report and they will fail, it will look like we failed to send the report although we did send it).

Plus Plus it is nice that all the information you need is under the same "summary" object. That way you don't have to think "Oh I ran send-report so I need to look at the final status".

monitor/cli.py
IDoneShaveIt: I agree with @elongl.

In addition, although summary will be sent through slack (for now), it doesn't mean it will always be like this, and that it is only relevant for sending the report to slack.
I can see the following flow:
run report -> upload to s3 -> notify on slack with the summary that the report is updated (and tag the relevant users).

I think that bundling between the report on slack to the summary is bad, and summary should be a feature by its own that we can combine with anything we would like (alerts + summary, slack report + summary, s3 + summary, only summary, etc).

clients/gcs/client.py
```diff
@@ -0,0 +1,35 @@
...
+logger = get_logger(__name__)
+
+
+class GCSClient:
+    def __init__(self, config: Config):
+        self.config = config
+        self.client = storage.Client.from_service_account_json(config.google_service_account_path)
+
+    @classmethod
+    def create_client(cls, config: Config) -> 'GCSClient':
```
IDoneShaveIt, line 18: Small one:
-> Optional[GCSClient]

clients/s3/client.py
```diff
@@ -0,0 +1,39 @@
...
+class S3Client:
+    def __init__(self, config: Config):
+        self.config = config
+        aws_session = boto3.Session(profile_name=config.aws_profile_name,
+                                    aws_access_key_id=config.aws_access_key_id,
+                                    aws_secret_access_key=config.aws_secret_access_key)
+        self.client = aws_session.client('s3')
+
+    @classmethod
+    def create_client(cls, config: Config) -> 'S3Client':
```
IDoneShaveIt, line 21: Small one:
-> Optional[S3Client

monitor/data_monitoring.py
```diff
@@ -160,23 +153,32 @@ def generate_report(self, tracking: AnonymousTracking, days_back: Optional[int]
...
+            if not send_succeded:
+                self.success = False
+
+        if self.gcs_client:
+            send_succeded = self.gcs_client.send_report(html_path)
+            self.execution_properties['sent_to_gcs_successfully'] = send_succeded
+            if not send_succeded:
+                self.success = False
+
+        self.execution_properties['success'] = self.success
```
IDoneShaveIt, line 181: I think you should still return self.success as well at the end of the method

IDoneShaveIt (approved):
Approved with small comments :)

monitor/data_monitoring.py
elongl: I do though.
```py
        return self.success
```

PR #193, opened 2022-08-09
Added documentation to Databricks support.

elongl:


docs/integrations/bigquery.mdx
```diff
@@ -11,7 +11,16 @@ Supports Bigquery since version 0.3.4
...
 ### BigQuery connection profile
 
-The profiles.yml file to connect to BigQuery should contain a profile in the following format:
+After installing Elementary's dbt package upon running `dbt deps`,
+you can generate Elementary's profile for usage with `edr` by running the following command within your project:
+
+```shell
+dbt run-operation elementary.generate_elementary_cli_profile
+```
+
```
Maayan-s, line 20: 1. Explain what will happen:
"The command will print to terminal..."

2. Explain the next step:
"copy the output to [HOME]/.dbt/profiles.yml, if it doe's not exist, here is a command to create it..."

3. What are the fields that are missing?

docs/integrations/bigquery.mdx
```diff
@@ -35,16 +44,19 @@ elementary:
       priority: interactive
 ```
 
-We support the same format and connection methods as dbt. Please refer to dbt's documentation of [BigQuery](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile) for further details.
+We support the same format and connection methods as dbt. Please refer to dbt's documentation
+of [BigQuery](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile) for further details.
 
-Note that although in dbt the 'location' parameter is optional, in Elementary lineage it is mandatory. Here are details about [BigQuery locations](https://cloud.google.com/bigquery/docs/locations).
+Note that although in dbt the 'location' parameter is optional, in Elementary lineage it is mandatory. Here are details
```
Maayan-s, line 50: I missed this, we can remove as the new docs don't include the lineage.

docs/integrations/databricks.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: 'Databricks'
+---
+
+## Partial Support
+
+Currently, Elementary only supports Dbt artifacts on Databricks.
+
```
Maayan-s, line 8: I don't think this explains it to the user. 
(Also, dbt is always lower case.) 

I would say:
"
Currently, Elementary Databricks integration is partial, and supports:
1. Installing the dbt package to upload dbt artifacts, test results and run results.
2. Data observability report generation.
3. Slack alerts on failures in models and tests. 

**Elementary anomaly detection tests are not supported yet**.
"

docs/integrations/bigquery.mdx
```diff
@@ -11,7 +11,16 @@ Supports Bigquery since version 0.3.4
 
 ### BigQuery connection profile
 
-The profiles.yml file to connect to BigQuery should contain a profile in the following format:
+After installing Elementary's dbt package upon running `dbt deps`,
```
Maayan-s, line 14: The cool thing about our migration to Mintlify is that docs are now code. 
You don't need to copy-paste any more. 

Create this part that explains the new profile generation command under the folder called 'snippets'.

Then on each page you want to use it, add this on the top:
`import GenerateElementaryProfile from "../snippets/generate-elementary-profile.mdx";`

And this where you want the actual snippet to be placed:
<GenerateElementaryProfile />

Maayan-s:
My overall feedback is that when you document new concepts that are completely unfamiliar to our users, like this macro, you should be as explicit as possible. 

Here is a good (funny) reference:
https://www.youtube.com/watch?v=Ct-lOOUqmyY

And a practical reference (look how explicit it is):
https://docs.getdbt.com/reference/commands/init

docs/integrations/databricks.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: 'Databricks'
+---
```
Maayan-s, line 3: This is a new page, so you need to add it to `mint.config.json`

Maayan-s:
Another comment - we have integrations in the front page as well, please add to there.

elongl:
> My overall feedback is that when you document new concepts that are completely unfamiliar to our users, like this macro, you should be as explicit as possible.
> 
> Here is a good (funny) reference: https://www.youtube.com/watch?v=Ct-lOOUqmyY
> 
> And a practical reference (look how explicit it is): https://docs.getdbt.com/reference/commands/init

Thanks a lot for the great feedback, I'll make sure to apply that in the future.

docs/integrations/bigquery.mdx
elongl: Thanks! I've added a custom link for each integration so making a generic snippet isn't as intuitive.
Is there a way to pass props to a snippet so I can do `<GenerateElementaryProfile profile_docs_link={...} />`?

PR #195, opened 2022-08-09
Support dimension anomalies

IDoneShaveIt:
Support for dimension anomalies in the report.
The only change that was required is to handle the test results samples for dimension anomalies in a different way from the rest.


monitor/dbt_project/macros/get_tests_sample_data.sql
```diff
@@ -40,12 +41,40 @@
...
+                {% set dimension_test_result_query %}
+                    select *
+                    from ({{test_results_query}})
+                    where is_anomalous
+                {% endset %}
+                {% set test_rows_sample = elementary_internal.get_test_rows_sample(dimension_test_result_query, test_type, metrics_sample_limit) %}
+                {% set anomalous = [] %}
+                {% for sample in test_rows_sample %}
+                    {% set anomalous_sample = {
+                        'end_time': sample['end_time'],
```
IDoneShaveIt, line 56: Currently the UI handles converting `end_time` to the right title.
This is important because our code breaks if it is not getting an `end_time`.
Because this is a temporary solution, I prefer to maintain this solution in one place (instead of handling it in multiple files) and that the UI will adjust the naming here.

monitor/dbt_project/macros/get_tests_sample_data.sql
```diff
@@ -40,12 +41,55 @@
...
+                    select *
+                    from ({{test_results_query}})
+                    where is_anomalous
+                {% endset %}
+                {% set test_rows_sample = elementary_internal.get_test_rows_sample(dimension_test_result_query, test_type, metrics_sample_limit) %}
+                {% set anomalous = [] %}
+                {% set headers = [{'id': 'elementary_end_time', 'display_name': 'timestamp', 'type': 'date'}] %}
+                {% for sample in test_rows_sample %}
+                    {% set anomalous_sample = {
+                        'elementary_end_time': sample['end_time'],
```
IDoneShaveIt, line 57: I use elementary prefix to prevent the edge case where the customers have dimensions that matches our columns

oravi (approved):

PR #201, opened 2022-08-09
Make BigQuery dataset explicitly refer to _elementary

Aylr:
This tiny PR makes the BigQuery quick start instructions more explicit. This avoids the empty report generated when erroneously configuring the dbt profile dataset to be `my_dataset` rather than `my_dataset_elementary`.


Maayan-s:
Hi @Aylr, thank you for making this contribution!

It's actually our first docs contribution, we just migrated so we could have the docs in the repo ([live now](https://docs.elementary-data.com/introduction)).
So this contribution makes me extremely happy!

I will add a comment in the code itself about the change.

docs/quickstart/generate-report-ui.mdx
```diff
@@ -72,7 +72,7 @@ elementary:
       keyfile: [full path to your keyfile]
 
       project: [project id]
-      dataset: [dataset name]
+      dataset: [dataset name]_elementary
```
Maayan-s, line 75: I agree that we need to make it more clear that this is the elementary dataset.
The issue is that we don't expect "[dataset name]_elementary", this is just a suggestion, you can change it if you want. It will even work if you don't configure a custom schema. 
We are [adding this macro](https://github.com/elementary-data/dbt-data-reliability/commit/55e68ef2f26e2eabf7a9a0f176df0b3456536b12) to the next release, to make it easier.

What I suggest is that you change:
`dataset: [dataset name]_elementary`
to:
`dataset: [dataset name] # elementary dataset, usually [dataset name]_elementary`

wdyt?

Aylr: @Maayan-s I'm delighted to be the first to exercise your new docs workflow!

Thanks for the suggestion. I've applied it to all three databases for consistency in experience.

Maayan-s:
Thank you!! Merging it now!

PR #228, opened 2022-08-15
Conditionally running 'dbt deps' in tracking initialization.

elongl:


monitor/dbt_project.py
```diff
@@ -0,0 +1,15 @@
+import os
```
IDoneShaveIt, line 1: I think I would rename the file to dbt_project_utils or something like that because now we have dbt_project directory and dbt_project file which is a bit confusing 🙂

tracking/anonymous_tracking.py
```diff
@@ -105,7 +105,9 @@ def track_cli_help(self):
 
     def _get_anonymous_warehouse_id(self):
         try:
-            dbt_runner = DbtRunner(monitor.paths.DBT_PROJECT_PATH, self.config.profiles_dir, self.config.profile_target)
+            dbt_runner = DbtRunner(monitor.dbt_project.PATH, self.config.profiles_dir, self.config.profile_target)
+            if not monitor.dbt_project.dbt_package_exists():
+                dbt_runner.deps()
```
IDoneShaveIt, line 110: what will happen if we won't be able to download the packages?
currently in the data_monitoring we handle it (raise an error and quit the run)

IDoneShaveIt: Plus I think that maybe a better way is to make sure we run it on the CLI.
Like before each command we should check for the deps.
That way we will have only one place that handles it and we won't have problems like that again in the future 🙂 
WDYT?

oravi (approved):

tracking/anonymous_tracking.py
elongl: I agree. This is a best-effort for a hotfix.
The codeblock is in a `try...except...` so if the `dbt deps` fails, worst-case it'll be re-handled in the `DataMonitoring`.
For now this is sufficient.

PR #230, opened 2022-08-15
Fix tracking

IDoneShaveIt:


clients/dbt/dbt_runner.py
```diff
@@ -35,20 +35,20 @@ def _run_command(
         if vars:
             json_vars = json.dumps(vars)
             dbt_command.extend(['--vars', json_vars])
-        if should_log:
+        if not quiet:
             logger.info(f"Running {' '.join(dbt_command)} (this might take a while)")
-        result = subprocess.run(dbt_command, check=False, capture_output=capture_output)
+        result = subprocess.run(dbt_command, check=False, capture_output=(json_output or quiet))
```
elongl, line 40:
```py
if quiet and not json_output:
        result = subprocess.run(dbt_command, check=False, stdout=subprocess.DEVNULL)
```

I think we should avoid capturing the data if not needed.

PR #239, opened 2022-08-15
Able to send alerts and report to private Slack channels.

elongl:
Tested on both alerts and report.
Remove the call for fetching all the channels.

### DIFF
--- a/clients/slack/client.py
+++ b/clients/slack/client.py
@@ -1,6 +1,5 @@
 import json
 from abc import ABC, abstractmethod
-from typing import List
 from typing import Optional
 
 from slack_sdk import WebClient, WebhookClient
@@ -52,13 +51,9 @@ class SlackWebClient(SlackClient):
         return WebClient(token=self.token)
 
     def send_message(self, channel_name: str, message: SlackMessageSchema, **kwargs) -> bool:
-        channel_id = self._get_channel_id(channel_name)
-        in_channel = self._join_channel(channel_id)
-        if not (channel_id and in_channel):
-            return False
         try:
             self.client.chat_postMessage(
-                channel=channel_id,
+                channel=channel_name,
                 text=message.text,
                 blocks=json.dumps(message.blocks) if message.blocks else None,
                 attachments=json.dumps(message.attachments) if message.attachments else None
@@ -69,13 +64,9 @@ class SlackWebClient(SlackClient):
             return False
 
     def send_file(self, channel_name: str, file_path: str, message: SlackMessageSchema) -> bool:
-        channel_id = self._get_channel_id(channel_name)
-        in_channel = self._join_channel(channel_id)
-        if not (channel_id and in_channel):
-            return False
         try:
             self.client.files_upload(
-                channels=channel_id,
+                channels=channel_name,
                 initial_comment=message.text,
                 file=file_path
             )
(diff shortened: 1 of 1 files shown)


clients/slack/client.py
```diff
@@ -124,7 +128,8 @@ def _get_channels(self) -> List[dict]:
             has_more = True
             cursor = None
             while has_more:
-                response = self.client.conversations_list(cursor=cursor)
+                response = self.client.conversations_list(cursor=cursor, types='public_channel,private_channel',
```
IDoneShaveIt, line 131: I think it will be better to leave the private_channel out of this.
Private channels should be handled manually by the user because they are sensitive.
If by accident the user gives extra permission to the app, I don't it to be able to join and post on private channels that it wasn't assigned to manually.

clients/slack/client.py
```diff
@@ -96,19 +101,18 @@ def send_report(self, channel_name: str, report_file_path: str):
...
-                channel
-                for channel
-                in channels
-                if channel["name"] == channel_name
-            ][0]
-            return channel["id"]
+            for channel in channels:
+                if channel['name'] == channel_name:
+                    return channel['id']
+            available_channel_names = [channel['name'] for channel in channels]
```
IDoneShaveIt, line 110: NIT:
you can create the `available_channel_names` in the first for loop (saves another run on the channels list)

clients/slack/client.py
```diff
@@ -52,30 +51,36 @@ def _initial_client(self):
...
+                channel_id = self._get_channel_id(channel_name)
+                if not channel_id:
+                    return False
+                joined_successfully = self._join_channel(channel_id=channel_id)
+                if not joined_successfully:
+                    return False
+                self.send_message(channel_name=channel_name, message=message)
+            elif err_type == 'channel_not_found':
+                logger.error(
+                    f'Channel was {channel_name} not found by the Elementary app. Please add the app to the channel.')
```
IDoneShaveIt, line 75:
```suggestion
                    f'Channel {channel_name} not found by the Elementary app. Please add the app to the channel.')
```

IDoneShaveIt (commented):
Like!

clients/slack/client.py
IDoneShaveIt: I'll explain myself better 🙂 
When you create a Slack application, you select the scopes of permissions you grants to it.
We currently don't guide the users to add permissions that allows the application to access private channels, but if by mistake they did grants those permissions (can be done very easily), it will be able to join any channel in the Slack.
So if by mistake the application has over privileges, it will have access to any private channel in the organizatoin (even if the user doesn't have).
This is an edge case, and mainly the user responsibility, but if we can avoid it on our side, I think we should.

The only meaning of this is if a user provides a name of a private channel, it will have to add the application manually to it first.
For public channels, it will join automatically.

elongl: image

elongl: We should also print the private channels that the app is a part of.
This can be useful for typos, etc.
Say there's a private channel called `data-ops` and the user wrote `dataops`.
If the list of available channels will not show `data-ops`, it will just be confusing to him and he will not assume that `data-ops` would work.

IDoneShaveIt: Agree!
For documentation sake - looking on Slack application permission scopes confirms that a Slack application can only see a private channels it has been added to.
Approving this solution 🙂

IDoneShaveIt (approved):
LGTM

PR #248, opened 2022-08-19
Debug logging (edr.log) a failure to load a module.

elongl:
Adds information to `edr.log` if loading a module fails.
Allows us to investigate the problem without requiring extra work from the user.


IDoneShaveIt (approved):
LGTM!

PR #249, opened 2022-08-19
Changed monitor's command title.

elongl:
In order to avoid confusion with the reports, `monitor` doesn't necessarily send Slack alerts.

image


IDoneShaveIt (approved):
LGTM!

PR #252, opened 2022-08-20
Not tracking purposeful exit.

elongl:


tracking/anonymous_tracking.py
```diff
@@ -83,6 +84,8 @@ def track_cli_end(self, module_name: str, execution_properties: dict, command: s
             pass
 
     def track_cli_exception(self, module_name: str, exc: Exception, command: str = None) -> None:
+        if isinstance(exc, click.exceptions.Exit):
```
IDoneShaveIt, line 87: mmmmm I am not sure about it 🤔 
What if someone ran a report, then canceled it in the middle? We will have a log for running the report, but no log that tells us that we don't have an end/fail log because the user canceled it in the middle

IDoneShaveIt (commented):
The code itself looks good, but I am not sure why we want to ignore those tracking?

tracking/anonymous_tracking.py
elongl: If he canceled, that'd be a `KeyboardInterrupt`.
A `click.exception.Exit` is an exception that is only raised by us (`ctx.exit(1)`).
This happens when a user provided invalid configuration.
Currently, `cli-exception` is considered an error or a bug on our side that we need to fix.

IDoneShaveIt: Got it. Thanks!

IDoneShaveIt (approved):
LGTM!

PR #259, opened 2022-08-22
Tracking fixes

elongl:


monitor/dbt_project/macros/get_adapter_type_and_unique_id.sql
```diff
@@ -1,5 +1,5 @@
-{% macro get_adapter_unique_id() %}
-    {{ elementary.edr_log(adapter.dispatch('get_adapter_unique_id')()) }}
+{% macro get_adapter_type_and_unique_id() %}
+    {{ elementary.edr_log(tojson([target.type, adapter.dispatch('get_adapter_unique_id')()])) }}
```
IDoneShaveIt, line 2: NIT:
Do we prefer list over dict?

tracking/anonymous_tracking.py
```diff
@@ -15,29 +18,35 @@
 logging.getLogger('posthog').disabled = True
 
 
+@dataclass
+class AnonymousWarehouse:
```
IDoneShaveIt, line 22: Let's use pydantic.
It is better to start and handle schemas in one way, and pydantic is much stronger and will better in the long run.

IDoneShaveIt: Plus you can move the hash method for the warehouse id into the schema.

tracking/anonymous_tracking.py
```diff
@@ -57,61 +66,61 @@ def init_anonymous_user_id(self):
...
-        if properties is None:
-            properties = dict()
+            if properties is None:
+                properties = dict()
 
-        properties['run_id'] = self.run_id
-        posthog.capture(distinct_id=self.anonymous_user_id, event=name,
-                        properties={'env': self.env_props, **properties},
-                        groups={'warehouse': self.anonymous_warehouse_id})
+            properties['run_id'] = self.run_id
```
IDoneShaveIt, line 76: I really don't like editing a dict we get as parameter to a method.
It can cause a lot of voodoo and a better practice IMO is to copy it and edit the copy
```python
extended_properties = {**properties, "run_id": self.run_id}
```

monitor/dbt_project/macros/get_adapter_type_and_unique_id.sql
elongl: Following the convention of `get_elementary_database_and_schema`.  
Also, this makes it possible to do `adapter_type, adapter_unique_id = self.dbt_runner.run_operation('get_adapter_type_and_unique_id')` which is kinda nice.

tracking/anonymous_tracking.py
elongl: Yeah, I definitely agree, btw this is not code I touched.

IDoneShaveIt (approved):

PR #270, opened 2022-08-24
Added 'elementary' profile validation.

elongl:


elementary/cli/cli.py
```diff
@@ -42,7 +42,7 @@ def get_command(self, ctx, name):
     def format_help(self, ctx, formatter):
         try:
             click.echo('Loading dependencies (this might take a few seconds)')
-            config = Config()
+            config = Config(should_validate_profile=False)
```
oravi, line 45: This is an indication that we shouldn't use the ctor for it. I suggest using config.validate_elementary_profile() in the CLI when we explicitly want to validate it. Also we can change it to return True / False and log inside the function a proper log based on the issue (file was not found or elementary profile was not found)

elementary/config/config.py
```diff
@@ -68,3 +75,12 @@ def has_aws(self) -> bool:
     @property
     def has_gcs(self):
         return self.gcs_bucket_name and self.google_service_account_path
+
+    def validate_elementary_profile(self):
+        profiles_path = os.path.join(self.profiles_dir, 'profiles.yml')
+        try:
+            profiles_yml = OrderedYaml().load(profiles_path)
```
oravi, line 82: I'd check explicitly if the path exists and if not return False, I'd return False if the elementary profile was not found

elongl: It's a more Pythonic approach.
_It's easier to ask for forgiveness than permission._

elementary/cli/cli.py
elongl: The thing is that in most situations you _would_ want to validate the profile.
It's safer to remove the validation in the single place in which you wouldn't want it (the help entry),
rather than if you were to add another command in which you would have to notice that you need to call the validate method and it wouldn't happen naturally.

elementary/config/config.py
elongl: If you returned a `boolean`, how would you know if it's not in `track-cli-exception`? 😉

oravi (approved):

PR #272, opened 2022-08-25
Updated FAQ

YashPimple:


Maayan-s:
Hi @YashPimple, 
Thank you for opening the PR. 
A couple of days ago I made some changes to the docs, and we now store FAQ as individual files (and later import to the main page). 

So first please create the file: `docs/snippets/faq/question-days-back.mdx`

In the file, we use a component called Accordion. 
The title is the question itself, and inside is the answer. 
I suggest something in the lines of:

```
<Accordion title="I changed `days back` but the results are the same, why is that?">

Elementary tests have a var named `days_back`. 
If you change it after executing elementary tests, you will need to run a full refresh to the metrics collected. This will make the next tests collect data for the new `days_back` timeframe.
The steps are: 

1. Change var `days_back` in your `dbt_project.yml`.
2. Full refresh of the model 'data_monitoring_metrics' by running ``` dbt run --select data_monitoring_metrics --full-refresh```.
3. Running the elementary tests again.

If you want the Elementary UI to show data for a longer period of time, use the days-back option of the CLI:
```edr monitor report --days-back 45```

</Accordion>
```

YashPimple:
@Maayan-s I have Push the respected PR as you have mentioned above.

PR #278, opened 2022-08-26
Attempting to automatically join on SlackClient.send_file().

elongl:
This is relevant if a user tries 'send-report' before 'monitor', in which case the application would have to be in the channel.


IDoneShaveIt (approved):
LGTM!

PR #284, opened 2022-08-28
support browser authentication method

IDoneShaveIt:
Our dbt_runner expect the log messages of dbt run-operation to be in json format.
However, browser authentication scenario returns log messages as string.
So instead of failing on json parsing, I log the failure and the unsupported log message to our edr file.
This is OK because run-operation log message should be in json format as expected, so if anything is returning as string it is currently OK to ignore it.
And if something will break in the future, we will have the edr log to understand what happened and what we should change.

### DIFF
--- a/elementary/clients/dbt/dbt_runner.py
+++ b/elementary/clients/dbt/dbt_runner.py
@@ -79,18 +79,22 @@ class DbtRunner:
         if log_errors and not success:
             logger.error(f'Failed to run macro: "{macro_name}"')
         run_operation_results = []
+        # breakpoint()
         if json_logs:
             json_messages = command_output.splitlines()
             for json_message in json_messages:
-                log_message_dict = json.loads(json_message)
-                log_message_data_dict = log_message_dict.get('data')
-                if log_message_data_dict is not None:
-                    if log_errors and log_message_dict['level'] == 'error':
-                        logger.error(log_message_data_dict)
-                        continue
-                    log_message = log_message_data_dict.get('msg')
-                    if log_message is not None and log_message.startswith(self.ELEMENTARY_LOG_PREFIX):
-                        run_operation_results.append(log_message.replace(self.ELEMENTARY_LOG_PREFIX, ''))
+                try:
+                    log_message_dict = json.loads(json_message)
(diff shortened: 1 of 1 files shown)


elongl:
I think this is a too broad way to handle this specific case, I really don't want to keep the `except Exception` trend in our code. Could you share the returned output when using browser authentication?

IDoneShaveIt:
@elongl can't reply your comment so starting a thread here.
I don't tend to agree in this case.
Starting to parse a specific log that might changed from platform to platform is not a good solution IMO.
This method parse dbt run-operation messages which should be in json format.
Skipping those who aren't and logging them to the edr will:
1. do the trick
2. not throw away useful information
3. not implement a specific support for an odd bug from dbt
4. break in the rest of the code if something will change the assumption of json format responses

Anyhow this is the message:
`Initiating login request with your identity provider. A browser window should have opened for you to complete the login. If you can't see it, check existing browser windows, or your OS settings. Press CTRL+C to abort and try again...`

We can have a specific handle for this message, I am just not sure how stable it is..

elongl:
1. Why would it change from platform to platform? It's a dbt thing.
2. Why not?
```py
if log.startswith('Initiating login request with your identity provider.'):
    logger.debug('Initiating authentication')
    continue
```

elongl:
How does it work? Does the `dbt` command just hold until the browser auth is complete?

IDoneShaveIt:
I will shed some light here:
1. This is message created in [snowflakedb/snowflake-connector-python](https://github.com/snowflakedb/snowflake-connector-python), making it very specific to snowflake (and I guess this is why they don't return it in the regular format)
2. The `dbt` command is waiting for the authentication to be complete before continue to run.
3. Trying to parse this message specific is possible, but as I said in point 1 it is very specific to snowflake.
4. Now when I except only JSONDecodeError, is it better, or still you are against any try except in this case?

If that a bog no no for you, I can go with the specific handling and fix it later in the worst case scenario

elongl:
I understand the problem better now. Thanks for elaborating. Approved.

elongl (approved):

PR #291, opened 2022-08-31
support uploading report to s3 folder

IDoneShaveIt:


elementary/clients/gcs/client.py
```diff
@@ -19,15 +19,23 @@ def __init__(self, config: Config):
...
         try:
             bucket = self.client.get_bucket(self.config.gcs_bucket_name)
-            blob = bucket.blob(report_filename)
+            blob = bucket.blob(bucket_report_path)
             blob.upload_from_filename(html_path, content_type='text/html')
-            logger.info('Uploaded report to GCS.')
+            logger.info('Uploaded report to GCS.')  
             if self.config.update_bucket_website:
-                bucket.copy_blob(blob, bucket, 'index.html')
+                bucket_report_filder_path = path.dirname(bucket_report_path)
```
oravi, line 33: Are you sure this will work on windows? I'm not sure it's a good practice to use path here because the bucket_report_path is not a local file system path. It's a URL / cloud based path.

elementary/clients/gcs/client.py
```diff
@@ -19,15 +19,23 @@ def __init__(self, config: Config):
     def create_client(cls, config: Config) -> Optional['GCSClient']:
         return cls(config) if config.has_gcs else None
 
-    def send_report(self, html_path: str) -> bool:
+    # html_path is the local path where the report generated at.
+    # bucket_file_path is where we want to upload the report to at the bucket - support folders!
+    def send_report(self, html_path: str, bucket_file_path: Optional[str] = None) -> bool:
```
oravi, line 24: I'd rename html_path to local_html_file_path and bucket_file_path to remote_bucket_file_path so it will be really clear in the code itself (then you can remove the comment above)

elementary/clients/gcs/client.py
```diff
@@ -19,15 +19,23 @@ def __init__(self, config: Config):
...
         report_filename = path.basename(html_path)
+        bucket_report_path = bucket_file_path if bucket_file_path else report_filename
```
oravi, line 26: here it's a bit confusing - why not saving it again to bucket_file_path?

elementary/clients/gcs/client.py
IDoneShaveIt: Dirname will work because it retrieves a path from the user.
If a user provides a path that S3 or GCS doesn't support, he should change it.
The only point I agree with is that the join should be done manually and not with path

elementary/clients/gcs/client.py
IDoneShaveIt: Override a variable that you get from the function argument is a bad practice IMO and always has the potential to cause weird behaviours.
There is the path you retrieve as an input, and there is the final report path that you actually provide to the bucket. Those are not always the same (depends on the case)

elementary/clients/s3/client.py
```diff
@@ -22,15 +22,19 @@ def __init__(self, config: Config):
     def create_client(cls, config: Config) -> Optional['S3Client']:
         return cls(config) if config.has_aws else None
 
-    def send_report(self, html_path: str) -> bool:
+    # html_path is the local path where the report generated at.
+    # bucket_file_path is where we want to upload the report to at the bucket - support folders!
+    def send_report(self, html_path: str, bucket_file_path: Optional[str] = None) -> bool:
         report_filename = path.basename(html_path)
```
oravi, line 28: Same comments as the GCS comments

elementary/clients/s3/client.py
```diff
@@ -22,15 +22,19 @@ def __init__(self, config: Config):
...
+        bucket_report_path = bucket_file_path if bucket_file_path else report_filename
         try:
-            self.client.upload_file(html_path, self.config.s3_bucket_name, report_filename,
+            self.client.upload_file(html_path, self.config.s3_bucket_name, bucket_report_path,
                                     ExtraArgs={'ContentType': 'text/html'})
             logger.info('Uploaded report to S3.')
             if self.config.update_bucket_website:
                 self.client.put_bucket_website(
                     Bucket=self.config.s3_bucket_name,
+                    # We use report_filename because a path can not be an IndexDocument Suffix.
```
oravi, line 37: Will the website work if we upload it to a path and here we use only the name itself?

elementary/monitor/data_monitoring.py
```diff
@@ -160,21 +160,21 @@ def generate_report(self, tracking: AnonymousTracking, days_back: Optional[int]
         self.execution_properties['success'] = self.success
         return self.success, html_path
 
-    def send_report(self, html_path: str) -> bool:
+    def send_report(self, html_path: str, host_file_path: Optional[str] = None) -> bool:
```
oravi, line 163: I'd rename to local_html_path and remote_file_path

elementary/monitor/cli.py
```diff
@@ -291,13 +292,16 @@ def send_report(
     anonymous_tracking.track_cli_start('monitor-send-report', get_cli_properties(), ctx.command.name)
     try:
         config.validate_send_report()
+        # If file name contains a folders (is a path),
+        # we want to make sure that we create the report localy without the folders of the full path.
+        local_file_path = os.path.basename(file_name) if file_name else None
```
oravi, line 297: Maybe we should add a new param called bucket_folder? then if you want to upload to bucket you can mention - 
bucket_name, bucket_folder and a file_name - wdyt?

IDoneShaveIt: I am not sure I am up to that idea..
Both S3 and GCS works in the way that you provide the path as part of the file name.
If you work with buckets, you should be familiar with that and I don't want to create a different behaviour.
Plus it can start getting weird that if you have folders hierarchy you provide a path under the bucket_folder and not just a folder name. In that case, what is the difference from passing it to the file name?

elementary/clients/gcs/client.py
IDoneShaveIt: Plus path on windows supports both / and \\ as separator

elementary/clients/s3/client.py
IDoneShaveIt: Yeah it creates an index that can be access via the bucket website (under the folder path)

docs/quickstart/share-report-ui.mdx
```diff
@@ -46,6 +49,12 @@ Alternatively, you can pass the access key ID and secret access key:
 edr monitor send-report --aws-access-key-id <ACCESS_KEY_ID> --aws-secret-access-key <SECRET_ACCESS_KEY> --s3-bucket-name <BUCKET_NAME>
 ```
 
+You can control the name of the report by providing `--bucket-file-path <REPORT_NAME>` to the `edr` command.
+The report will be uploaded to your bucket under the provided name:
```
oravi, line 53: under the provided path

docs/quickstart/share-report-ui.mdx
```diff
@@ -46,6 +49,12 @@ Alternatively, you can pass the access key ID and secret access key:
...
+
+- Make sure the report name you provided has a suffix of `.html`.
```
oravi, line 55: Make sure the bucket file path ends with a suffix of '.html'

docs/quickstart/share-report-ui.mdx
```diff
@@ -46,6 +49,12 @@ Alternatively, you can pass the access key ID and secret access key:
...
+- If you wish to upload the report into a folder in your bucket, you should add the folder name to the provided report name: `folder/report_name.html`
```
oravi, line 56: This is no longer relevant

docs/quickstart/share-report-ui.mdx
```diff
@@ -14,16 +14,19 @@ After you [setup a Slack app and token](../integrations/slack#slack-integration-
 edr monitor send-report --slack-token <SLACK_TOKEN> --slack-channel-name <CHANNEL_NAME>
 ```
 
-A report will be sent to your Slack channel.
+You can control the name of the report by providing `--slack-file-name <REPORT_NAME>` to the `edr` command.
```
oravi, line 17: to edr monitor send-report

docs/quickstart/share-report-ui.mdx
```diff
@@ -79,6 +88,12 @@ Once you have it, you can send the report to your GCS bucket by running the comm
 edr monitor send-report --google-service-account-path <SERVICE_ACCOUNT_PATH> --gcs-bucket-name <BUCKET_NAME>
 ```
 
+You can control the name of the report by providing `--bucket-file-path <REPORT_NAME>` to the `edr` command.
```
oravi, line 91: See docs comments above

elementary/clients/gcs/client.py
oravi: Maybe it will work  but it's not a good practice as it's a URL. Why not being more explicit? like simply splitting by '/' as you've done below? also there's a typo there

elementary/monitor/cli.py
```diff
@@ -291,13 +300,16 @@ def send_report(
     anonymous_tracking.track_cli_start('monitor-send-report', get_cli_properties(), ctx.command.name)
     try:
         config.validate_send_report()
+        # If file name contains a folders (is a path),
+        # we want to make sure that we create the report localy without the folders of the full path.
+        local_file_path = os.path.basename(bucket_file_path) if bucket_file_path else slack_file_name
```
oravi, line 305: Same here - it's a not a local folder. I'd suggest doing something like - bucket_file_path.rsplit('/', 1)

elementary/monitor/cli.py
```diff
@@ -291,13 +300,16 @@ def send_report(
     anonymous_tracking.track_cli_start('monitor-send-report', get_cli_properties(), ctx.command.name)
     try:
         config.validate_send_report()
+        # If file name contains a folders (is a path),
+        # we want to make sure that we create the report localy without the folders of the full path.
```
oravi, line 304: Update comment to something like - "bucket-file-path determines the path of the report in the bucket. If this path contains folders we extract the report file name to first save the report locally"

elementary/clients/utils/bucket_path.py
```diff
@@ -0,0 +1,19 @@
...
+BUCKET_PATH_SEP = '/'
+
+
+def dirname(bucket_path: str) -> Optional[str]:
+    dir_path = BUCKET_PATH_SEP.join(bucket_path.split(BUCKET_PATH_SEP)[:-1])
+    return dir_path if dir_path else None
+
+
+def basename(bucket_path: str) -> str:
+    return bucket_path.rsplit(BUCKET_PATH_SEP, 1)[-1]
```
oravi, line 12: you don't need [-1] if you use rsplit(sep, 1)

elementary/clients/s3/client.py
```diff
@@ -22,15 +22,17 @@ def __init__(self, config: Config):
     def create_client(cls, config: Config) -> Optional['S3Client']:
         return cls(config) if config.has_aws else None
 
-    def send_report(self, html_path: str) -> bool:
-        report_filename = path.basename(html_path)
+    def send_report(self, local_html_file_path: str, remote_bucket_file_path: Optional[str] = None) -> bool:
+        report_filename = basename(local_html_file_path)
```
oravi, line 26: Here you work with the local file so here you do need to work with Path

elementary/clients/gcs/client.py
```diff
@@ -19,15 +19,21 @@ def __init__(self, config: Config):
...
             bucket = self.client.get_bucket(self.config.gcs_bucket_name)
-            blob = bucket.blob(report_filename)
-            blob.upload_from_filename(html_path, content_type='text/html')
-            logger.info('Uploaded report to GCS.')
+            blob = bucket.blob(bucket_report_path)
+            blob.upload_from_filename(local_html_file_path, content_type='text/html')
+            logger.info('Uploaded report to GCS.')  
             if self.config.update_bucket_website:
-                bucket.copy_blob(blob, bucket, 'index.html')
+                bucket_report_filder_path = path.dirname(bucket_report_path)
```
oravi, line 31: Here you are still using path.dirname?

elementary/clients/utils/bucket_path.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def join_path(path_parts: list) -> str:
+    if len(path_parts) == 1:
```
oravi: you don't need this, the join will work even if list has length of one

elementary/monitor/cli.py
oravi: Still using path.basename

PR #292, opened 2022-08-31
Remove elementary models from the report

IDoneShaveIt:
Currently we show elementary models (artifacts) in our report.
Those models get created in elementary package which the customers has no "control" on it, meaning that they can't add any test for them - causing them be redundant in the report (and they may confuse users as well).

In this PR I suggest to filter them out from our report.


elementary/monitor/dbt_project/macros/artifacts_queries/dbt_models_query.sql
```diff
@@ -0,0 +1,9 @@
+{% macro dbt_models_query() %}
```
oravi, line 1: My intuition is that in this case as it's simply filtering the package maybe it's better to just have the package != elementary in all 4 places?

PR #294, opened 2022-09-01
Alerting on slack when there is no alerts to be send

### LINKED ISSUE #133: Add "test alerting" on Slack
Slack alerts are only sent if there are failed tests.
When users deploy Elementary, they want to have a way to validate the deployment worked. 
We should have a flag for validating the deployment.

hanrok:
Resolves #133 
* Send test message on the CLI when there is no other alert to send on slack

## TBD
* Add/Do tests to understand it works correctly

### DIFF
--- a/elementary/monitor/cli.py
+++ b/elementary/monitor/cli.py
@@ -101,6 +101,12 @@ def get_cli_properties() -> dict:
     default=None,
     help="Specify raw YAML string of your dbt variables."
 )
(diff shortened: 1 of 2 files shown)


elongl:
This was merged [here](https://github.com/elementary-data/elementary/commit/97e92bd57f5d95744dcfbe1703521ea8b5f5c3f1).

PR #303, opened 2022-09-06
Support alert slack channel from test level

IDoneShaveIt:
We now support configuring a slack channel for alerts at the test level (Adding a channel argument to a test meta field).
If a channel is configured both at the test level and the model level, we use the test level channel.

Also fixed rate limitation bug with slack!


elementary/clients/slack/client.py
```diff
@@ -130,7 +131,11 @@ def _get_channels(self) -> List[dict]:
 
     def _handle_send_err(self, err: SlackApiError, channel_name: str) -> bool:
         err_type = err.response.data['error']
-        if err_type == 'not_in_channel':
+        if err_type == 'ratelimited':
+            rate_limit_waiting_time = int(err.response.headers['Retry-After'])
```
oravi, line 135: add try-except here so we won't fail on trying to parse their retry-after? Also add log that we got rate limit from Slack and therefore waiting X sec (X should be what you get from Slack)

oravi: Also - can we get rate limit also if a user uses webhook? how should we handle it?

docs/quickstart/send-slack-alerts.mdx
```diff
@@ -154,8 +153,8 @@ models:
 
 <Accordion title="Custom Slack channel">
 By default Elementary uses the Slack channel that was configured in the Slack integration.
-Elementary supports configuring custom slack channels that are configured on your models / sources and snapshots.
-If you configure a custom slack channel for a model, all the test alerts that belong to this model will be sent to this custom slack channel as well.
+Elementary supports configuring custom slack channels that are configured on your models / sources / tests and snapshots.
+If you configure a custom slack channel for a model, all the test alerts that belong to this model will be sent to this custom slack channel.
```
oravi, line 158: Also add - if you configure both a channel for a model and a channel for a test - the channel for a test will be used

docs/quickstart/send-slack-alerts.mdx
```diff
@@ -174,9 +173,24 @@ If your models are in folders by department / team, another useful option is to
...
+You can also configure a custom slack channel for a specific test.
+
+```yml properties.yml
+models:
+  - name: marketing_leads
+    columns:
+      - name: column_name
+        tests:
+          - unique:
+              meta:
```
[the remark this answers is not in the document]
IDoneShaveIt: directly here under meta :)

elementary/clients/slack/client.py
IDoneShaveIt: You are right about the webhook. Added handler for it as well.

IDoneShaveIt: I tested it and it seems like there is no rate limit on webhooks.
The error format is different from token and if I can't raise a limit error for webhooks, I can't create the right way to handle it..

docs/quickstart/send-slack-alerts.mdx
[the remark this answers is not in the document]
oravi: Will it work if a user will do it under - config: meta: channel_name?

elementary/clients/slack/client.py
oravi: It seems from the docs that you can get a rate limit also to webhooks - 
image

https://api.slack.com/docs/rate-limits

Seems like we should check the http error code and the headers

elementary/clients/slack/client.py
```diff
@@ -130,7 +131,12 @@ def _get_channels(self) -> List[dict]:
 
     def _handle_send_err(self, err: SlackApiError, channel_name: str) -> bool:
         err_type = err.response.data['error']
-        if err_type == 'not_in_channel':
+        if err_type == 'ratelimited':
```
oravi, line 134: Seems like you can get rate limit also for webhooks - 
image


I think we should check the http error code and extract the retry after from the headers

IDoneShaveIt: They are saying you can, but I run it with an endless loop and got none.
I know that we can try and extract the header but the response is in a different format, and I am not sure how to extract it because I haven't been able to produce this error.

oravi: you don't need to reproduce - it's simply reading the status code and extracting it from the headers in the response (according to the spec in their docs)

oravi (approved):

PR #306, opened 2022-09-12
Use test results view instead of elementary test results table on tests API

IDoneShaveIt:
Use test results view instead of elementary test results table on tests API


elongl (commented):
Looks good, just do the renaming according to the other PR.

elongl (approved):

PR #319, opened 2022-09-19
add model runs to elementary report

IDoneShaveIt:
- Added new methods to models API to support model runs page
- Added model runs to the elementary report json


elementary/monitor/api/models/models.py
```diff
@@ -1,16 +1,86 @@
...
+                    status=model_run['status'],
+                    full_refresh=model_run['full_refresh'],
+                    materialization=model_run['materialization'],
+                    execution_time=model_run['execution_time']
+                )
+                for model_run
+                in model_runs
+            ]
+            # We want to keep failed model runs out from the median.
+            execution_times_for_median = [model_run['execution_time'] for model_run in model_runs if model_run['status'] != 'error']
```
oravi, line 55: I think we should exclude skipped models as well so I think a better filtering could be - 
`if lower(model_runs['status']) == 'success'`

elementary/monitor/api/models/models.py
```diff
@@ -1,16 +1,86 @@
...
+                    materialization=model_run['materialization'],
+                    execution_time=model_run['execution_time']
+                )
+                for model_run
+                in model_runs
+            ]
+            # The median should be based only on succesfull model runs.
+            successful_execution_times = [model_run['execution_time'] for model_run in model_runs if model_run['status'].lower() == 'success']
+            median_execution_time = statistics.median(successful_execution_times) if len(successful_execution_times) else 0
+            last_model_run = sorted(model_runs, key=lambda run: run['generated_at'])[-1]
```
oravi, line 57: This will fail if model_runs is empty - also maybe worth doing it in the db so it will be faster?

oravi (approved):

PR #325, opened 2022-09-24
Output folder flag for Elementary CLI

### LINKED ISSUE #313: Output folder flag for Elementary CLI
Keeping artifacts of elementary runs (edr.log, json files) in a separate directory, mainly to help with git and folders clutter.

eran-levy:
resolves #313

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ var/
 *.egg
 *.mypy_cache/
 logs/
+output/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ var/
 *.egg
 *.mypy_cache/
 logs/
-output/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -14,7 +14,7 @@ class Config:
 
     DEFAULT_CONFIG_DIR = str(Path.home() / '.edr')
     DEFAULT_PROFILES_DIR = str(Path.home() / '.dbt')
-    DEFAULT_TARGET_DIR = 'output'
+    DEFAULT_TARGET_DIR = 'edr_output'
 
     def __init__(self, config_dir: str = DEFAULT_CONFIG_DIR, profiles_dir: str = DEFAULT_PROFILES_DIR,
                  profile_target: str = None, update_bucket_website: bool = None, slack_webhook: str = None,
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -13,18 +13,20 @@ class Config:
 
     DEFAULT_CONFIG_DIR = str(Path.home() / '.edr')
     DEFAULT_PROFILES_DIR = str(Path.home() / '.dbt')
+    DEFAULT_TARGET_DIR = 'output'
 
     def __init__(self, config_dir: str = DEFAULT_CONFIG_DIR, profiles_dir: str = DEFAULT_PROFILES_DIR,
                  profile_target: str = None, update_bucket_website: bool = None, slack_webhook: str = None,
                  slack_token: str = None, slack_channel_name: str = None, aws_profile_name: str = None,
                  aws_access_key_id: str = None, aws_secret_access_key: str = None, s3_bucket_name: str = None,
-                 google_service_account_path: str = None, gcs_bucket_name: str = None):
+                 google_service_account_path: str = None, gcs_bucket_name: str = None,
+                 target_dir: str = DEFAULT_TARGET_DIR):
         self.config_dir = config_dir
         self.profiles_dir = profiles_dir
         self.profile_target = profile_target
         config = self._load_configuration()
 
-        self.target_dir = config.get('target-path') or os.getcwd()
+        self.target_dir = config.get('target-path') or target_dir
 
         self.update_bucket_website = update_bucket_website or config.get('update_bucket_website', False)
 
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -1,5 +1,6 @@
 import os
 from pathlib import Path
+from dateutil import tz
 
 from elementary.exceptions.exceptions import NoElementaryProfileError, NoProfilesFileError, InvalidArgumentsError
 from elementary.utils.ordered_yaml import OrderedYaml
@@ -17,7 +18,7 @@ class Config:
 
     def __init__(self, config_dir: str = DEFAULT_CONFIG_DIR, profiles_dir: str = DEFAULT_PROFILES_DIR,
                  profile_target: str = None, update_bucket_website: bool = None, slack_webhook: str = None,
-                 slack_token: str = None, slack_channel_name: str = None, aws_profile_name: str = None,
+                 slack_token: str = None, slack_channel_name: str = None, timezone: str = None, aws_profile_name: str = None,
                  aws_access_key_id: str = None, aws_secret_access_key: str = None, s3_bucket_name: str = None,
                  google_service_account_path: str = None, gcs_bucket_name: str = None,
                  target_dir: str = DEFAULT_TARGET_DIR):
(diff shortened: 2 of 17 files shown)


elementary/config/config.py
```diff
@@ -13,18 +13,20 @@ class Config:
 
     DEFAULT_CONFIG_DIR = str(Path.home() / '.edr')
     DEFAULT_PROFILES_DIR = str(Path.home() / '.dbt')
+    DEFAULT_TARGET_DIR = 'output'
```
elongl, line 16: I believe this should be called `edr_output`.
Given that `edr` can be run from every directory,
it might be confusing to just have the `output` directory.

elementary/config/config.py
```diff
@@ -13,18 +13,20 @@ class Config:
...
-                 google_service_account_path: str = None, gcs_bucket_name: str = None):
+                 google_service_account_path: str = None, gcs_bucket_name: str = None,
+                 target_dir: str = DEFAULT_TARGET_DIR):
         self.config_dir = config_dir
         self.profiles_dir = profiles_dir
         self.profile_target = profile_target
         config = self._load_configuration()
 
-        self.target_dir = config.get('target-path') or os.getcwd()
+        self.target_dir = config.get('target-path') or target_dir
```
elongl, line 29: I tend to agree that it would be a good idea to put our outputs under a single directory by default rather than the working directory, but we'll need to discuss this change as users might currently rely on the path of the report and such.

elementary/monitor/cli.py
```diff
@@ -166,15 +166,22 @@ def monitor(
     '--disable-passed-test-metrics',
     type=bool,
     default=False,
-    help="If set to true elementary report won't show data metrics for passed tests (this can improve report creation time)."
+    help="If set to true elementary report won't show data metrics for passed tests (this can improve report "
+         "creation time)."
+)
+@click.option(
```
elongl, line 172: We have the `common_options` function above that is designed to aggregate CLI options that are received on all the `monitor` commands.

Could you please move the `click.option` to there?

elementary/lineage/cli.py
```diff
@@ -127,15 +127,21 @@ def get_cli_lineage_generate_properties() -> dict:
     cls=RequiredIf,
     required_if='profiles_dir'
 )
+@click.option(
```
elongl, line 130: Sorry for asking this, could you please remove all the changes under `elementary/lineage`?
The reason is that this module is deprecated and we'll delete it soon.
Given that, introducing more changes to it will just increase the likelihood of breaking things.

elementary/utils/log.py
```diff
@@ -1,10 +1,11 @@
 import logging
+import os.path
 import sys
 
 from elementary.utils.env_vars import is_debug_mode_on
 
 FORMATTER = logging.Formatter("%(asctime)s — %(name)s — %(levelname)s — %(message)s")
-LOG_FILE = "edr.log"
+LOG_FILE = "logs/edr.log"
```
elongl, line 8: I believe this should also use the `target-dir`, though I can see how this is a bit more challenging.

elongl (changes requested):
Looks awesome! Thank you very much for the contribution.
Left a couple of comments.

elementary/config/config.py
eran-levy: Agree, you're right

elementary/config/config.py
eran-levy: Correct, lmk what you think, in general I assume the flat working dir might giving the user headache anyway as they might move and cleanup.. we can also use the os.getcwd() as the default

elementary/monitor/cli.py
eran-levy: At the beginning I placed it in the common_options but then I realized that its not really used in the "monitor" command so I thought it might look confusing and changed. 
No prob, moving to common_options

elementary/lineage/cli.py
eran-levy: done

elementary/utils/log.py
eran-levy: Since logs exist also for dbt, I recommend it to be "logs" as well. Since we are logging while initializing the cli in elementary/cli/cli.py utilizing the taget_dir option on init is bit problematic. What do you suggest?

oravi: @eran-levy @elongl what do you think about maybe adding the file handler to the logger only after we initialize the config object? (the logger can be initialized with only a console handler by default)

elementary/monitor/cli.py
```diff
@@ -46,7 +47,13 @@ def common_options(func):
         type=str,
         default=Config.DEFAULT_PROFILES_DIR,
         help="Specify your profiles dir where a profiles.yml is located, this could be a dbt profiles dir "
-             "(if your profiles dir is ~/.dbt, no need to provide this parameter as we use it as default).",
+             "(if your profiles dir is ~/.dbt, no need to provide this parameter as we use it as default)."
+    )(func)
+    func = click.option(
+        '--target-dir', '-td',
```
oravi, line 53: Maybe let's use `--target-path` to be consistent with the flag in config.yml?

elementary/config/config.py
oravi: Agree with both of you :) maybe we can start with default as os.getcwd(). Then we can announce in the community that we are going to change the default from cwd to 'edr_output' and deploy this change afterwards?

elementary/monitor/data_monitoring.py
```diff
@@ -130,6 +130,7 @@ def generate_report(self, tracking: AnonymousTracking, days_back: Optional[int]
         bool, str]:
         now_utc = get_now_utc_str()
         html_path = self._get_report_file_path(now_utc, file_path)
+        os.makedirs(os.path.dirname(html_path), exist_ok=True)
```
oravi, line 133: What do you think about using `makedirs` in the config object? this way if we add more files to the target dir we don't need to remember calling this code in every flow

.gitignore
```diff
@@ -26,6 +26,7 @@ var/
 *.egg
 *.mypy_cache/
 logs/
+output/
```
oravi, line 29: If you keep the default `edr_output`, is it supposed to be `edr_output`?

oravi:
@eran-levy thanks so much for opening a PR!

.gitignore
eran-levy: Good catch missed it, done

elementary/utils/log.py
eran-levy: @oravi Its possible but I think less recommended from design perspective as it will require that every new subcommand that will be added to remember and init the log file handler as well. Tight coupling both functionalities can be a bit problematic in my opinion.

elementary/monitor/cli.py
eran-levy: Agree, done

elementary/config/config.py
eran-levy: Yup, good point

elementary/monitor/data_monitoring.py
eran-levy: Good point, done

elementary/utils/log.py
oravi: @eran-levy I agree that adding it to every subcommand is not good but maybe there's a way around it. Try to see how the CLI commands are structured and if you can maybe implement it in a single command

.gitignore
eran-levy: Since we dont have edr_output default anymore and its going to be os.getcwd(), so removed that

eran-levy:
@oravi just pushed them now

elementary/utils/log.py
eran-levy: ok checking

eran-levy: @oravi I couldnt find an elegant way to abstract the creation of the logger and config while running subcommand - before moving forward so we wont write throw-away code, IMHO there are 2 options here:
1. Just adding a logger handler creation on every subcommand (in monitor command because I understood from @elongl lineage is going to be deprecated which means that some subcommands will have the functionality and some not) tightly coupled with the config and given arguments - I wont be able to add to all and if the user won't provide an argument we will need to store it in some default folder anyway.
2. Continue with the current approach - just persist it under "logs" (I think preferable as this is what dbt uses) or in another folder such as: "edr_output"

what do you suggest?

PR #327, opened 2022-09-25
Added support for OAuth authentication for GCloud.

elongl:


elementary/clients/gcs/client.py
```diff
@@ -2,10 +2,12 @@
 from typing import Optional
 
 import google
+from google.auth.credentials import Credentials
 from google.cloud import storage
+from google.oauth2 import service_account
```
oravi, line 7: Should we add any of these imports to setup.py?

elongl: No, they're all included in `google-cloud-storage`.
You can tell this is the case due to the tests working properly.

oravi (approved):

PR #347, opened 2022-10-03
Added a flag that excludes Elementary's models.

elongl:


elongl:
[This](https://github.com/elementary-data/dbt-data-reliability/pull/114) is a requirement.

Maayan-s:
Don't you think it should be true by default?

elongl:
I think that most users would want to be able to look at our own lineage when trying to figure out which tables contain information they might be looking for and for understanding how we generally work.

elementary/monitor/cli.py
```diff
@@ -213,6 +213,12 @@ def monitor(
     default=True,
     help="Whether to open the report in the browser.",
 )
+@click.option(
+    "--exclude-elementary-models",
+    type=bool,
+    default=False,
+    help="Exclude Elementary's own models in the report.",
```
oravi, line 220: cosmetics - "Exclude Elementary's internal models from the report"

PR #351, opened 2022-10-03
Source freshness

elongl:


elementary/monitor/dbt_project/macros/get_new_source_freshness_alerts.sql
```diff
@@ -0,0 +1,45 @@
+{% macro get_new_source_freshness_alerts(days_back) %}
+    -- depends_on: {{ ref('alerts_source_freshness') }}
+    {% set elementary_database, elementary_schema = elementary.target_database(), target.schema %}
```
oravi, line 3: Where do you use this?

elementary/monitor/dbt_project/macros/get_new_source_freshness_alerts.sql
```diff
@@ -0,0 +1,45 @@
...
+            select * from {{ ref('alerts_source_freshness') }}
+            where alert_sent = false and {{ elementary.cast_as_timestamp('detected_at') }} >= {{ get_alerts_time_limit(days_back) }}
+        ),
+        sources as (
+            select * from {{ ref('elementary', 'dbt_sources') }}
+        )
+
+        select *
+        from new_alerts
+        join sources on new_alerts.unique_id = sources .unique_id
```
oravi, line 16: why do you need this join if alerts_source_freshness already joins the sources to the source freshness results?

elementary/utils/time.py
```diff
@@ -39,3 +39,10 @@ def format_milliseconds(duration: int) -> str:
     )
 
     return f"{hours}h:{minutes}m:{seconds}s:{remaining_milliseconds}ms"
+
+
+def try_parse_time(isoformat_datetime: str) -> str:
```
oravi, line 44: WDYT about renaming to convert_isoformat_to_custom_format and receive the format as default DATETIME_FORMAT?

elementary/monitor/alerts/source_freshness.py
```diff
@@ -0,0 +1,101 @@
...
+        path: str,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.unique_id = unique_id
+        self.snapshotted_at = try_parse_time(snapshotted_at) if snapshotted_at else None
+        self.max_loaded_at = try_parse_time(max_loaded_at) if max_loaded_at else None
+        self.max_loaded_at_time_ago_in_s = max_loaded_at_time_ago_in_s
+        self.source_name = source_name
+        self.identifier = identifier
```
oravi, line 35: Do you know when identifier exists for a source and when name exists? I wonder what's the difference between them

elementary/monitor/alerts/source_freshness.py
```diff
@@ -0,0 +1,101 @@
...
+        freshness_error_after: str,
+        freshness_warn_after: str,
+        freshness_filter: str,
+        path: str,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.unique_id = unique_id
+        self.snapshotted_at = try_parse_time(snapshotted_at) if snapshotted_at else None
+        self.max_loaded_at = try_parse_time(max_loaded_at) if max_loaded_at else None
```
oravi, line 32: Do we need to convert these to the local / received timezone?

elementary/monitor/alerts/source_freshness.py
```diff
@@ -0,0 +1,101 @@
...
+        self.max_loaded_at_time_ago_in_s = max_loaded_at_time_ago_in_s
+        self.source_name = source_name
+        self.identifier = identifier
+        self.freshness_error_after = freshness_error_after
+        self.freshness_warn_after = freshness_warn_after
+        self.freshness_filter = freshness_filter
+        self.path = path
+
+    def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
```
oravi, line 41: Alert formatting looks good to me, can you send some examples in the snap channel to hear Talmi's feedback?

elementary/monitor/alerts/source_freshness.py
elongl: image

From what I'm seeing throughout most of the code,
it seems like `name` is either just a wrapper for `identifier`, or that it was deprecated for it.
Anyhow, from what I've sent, you can see that `identifier` is always present as it fallbacks to `name`.

elementary/monitor/alerts/source_freshness.py
elongl: Already did, sent him in private a couple of days ago.

elementary/monitor/alerts/source_freshness.py
elongl: Great catch! Added.

elementary/monitor/dbt_project/macros/get_new_source_freshness_alerts.sql
elongl: Thanks! That was before I did it in the alerts model.

elementary/monitor/dbt_project/models/alerts/alerts_source_freshness.sql
```diff
@@ -0,0 +1,19 @@
...
+  config(
+    materialized = 'incremental',
+    unique_key = 'alert_id',
+    merge_update_columns = ['alert_id'],
+    on_schema_change = 'append_new_columns'
+  )
+}}
+
+
+{% set error_source_freshness_relation = adapter.get_relation(this.database, this.schema, 'alerts_dbt_source_freshness') %}
```
oravi, line 11: cosmetics - the name here is confusing IMO because when I read it I think that this is a relation contains errors in source freshness and not alerts

elementary/monitor/dbt_project/macros/empty_alert_tables.sql
```diff
@@ -1,3 +1,7 @@
 {% macro empty_alerts_models() %}
 	{{ elementary.empty_table([('alert_id', 'string'), ('unique_id', 'string'), ('detected_at', 'string'), ('database_name', 'string'), ('materialization', 'string'), ('path', 'string'), ('schema_name', 'string'), ('message', 'string'), ('owners', 'string'), ('tags', 'string'), ('alias', 'string'), ('status', 'string'), ('full_refresh', 'boolean'), ('alert_sent', 'boolean'), ('original_path', 'string')]) }}
 {% endmacro %}
+
+{% macro empty_alerts_source_freshness() %}
```
oravi, line 5: The schema here doesn't match the real schema of the table when there are alerts for source freshness

elementary/monitor/dbt_project/models/alerts/alerts_source_freshness.sql
elongl: Agreed. Copied it from `alerts_models` but there the alerts are actually errors so it's different.

PR #363, opened 2022-10-23
Checking compatibility between edr and dbt package version.

elongl:


elementary/utils/package.py
```diff
@@ -13,8 +17,19 @@ def get_package_version() -> Optional[str]:
...
+
+
+def check_dbt_pkg_compatible(dbt_pkg_ver: str):
+    dbt_pkg_ver = version.parse(dbt_pkg_ver)
+    py_pkg_ver = version.parse(get_package_version())
+    if dbt_pkg_ver.major != py_pkg_ver.major or dbt_pkg_ver.minor != py_pkg_ver.minor:
+        click.secho(
+            "You are using incompatible versions between 'edr' and Elementary's dbt package.\n"
+            "Please change the major and minor versions to align.\n",
+            fg="yellow",
```
oravi, line 34: Maybe print that minor and major should be aligned so the user will know what to do?

elementary/utils/package.py
```diff
@@ -13,8 +17,19 @@ def get_package_version() -> Optional[str]:
...
         resp.raise_for_status()
         return resp.json().get("info", {}).get("version")
     except Exception:
         return None
+
+
+def check_dbt_pkg_compatible(dbt_pkg_ver: str):
+    dbt_pkg_ver = version.parse(dbt_pkg_ver)
+    py_pkg_ver = version.parse(get_package_version())
+    if dbt_pkg_ver.major != py_pkg_ver.major or dbt_pkg_ver.minor != py_pkg_ver.minor:
```
oravi, line 30: What do you think about printing that the versions are compatible?

elongl: I'll make a debug log as I'm printing that I'm doing this check.

PR #366, opened 2022-10-26
Tracking Elementary Hosted.

elongl:


elongl:
Here's the [addition of it](https://github.com/elementary-data/elementary-internal/pull/40/files#r1004545936).

PR #372, opened 2022-10-30
Unify report times formats

IDoneShaveIt:
Making all the times in our system to be at the same time format - ISO. format without microseconds.
Plus added performance improvements of the UI.


elementary/monitor/alerts/alert.py
```diff
@@ -31,13 +32,15 @@ def __init__(
         self.elementary_database_and_schema = elementary_database_and_schema
         self.detected_at_utc = None
         self.detected_at = None
+        self.detected_at_utc_str = None
```
elongl, line 35: What do you think about having a single `detected_at` which is a `datetime` object with timezone included and format it whenever needed? Seems a bit exaggerated to have 4 different class properties which are roughly the same thing.

elementary/utils/time.py
```diff
@@ -51,3 +55,33 @@ def convert_datetime_utc_str_to_timezone_str(
         )
     except Exception:
         return isoformat_datetime
+
+
+def convert_partial_utc_iso_format_to_full_iso_format(
```
elongl, line 60: These functions are very similar.
What do you think about having an `is_utc` flag and merge them, or alternatively set default timezone to UTC if no timezone is defined?

elongl (changes requested):
Looks good, left two small comments.

elementary/monitor/alerts/alert.py
IDoneShaveIt: You are right.

elementary/utils/time.py
IDoneShaveIt: Sounds good.

PR #381, opened 2022-11-03
evaluate config args in hierarchy, even the "falsy" ones

### LINKED ISSUE #336: Configuration does not properly handle precedence (CLI Arguments -> Config File -> Default)
The current lookup for configuration in the [`Config`](https://github.com/elementary-data/elementary/blob/master/elementary/config/config.py) object is based on the value's truthiness.
For instance: `self.conf_val = cli_val or file_conf.get('my_val', True)`.

If the `cli_val` was specified `False` and not `None`, the default `True` of the configuration value would be used regardless because the expression evaluates to `False or True` which is `True`.

A simple solution will be to introduce a `_lookup()` method that checks according to the precedence that values are not `None`.

```py
def _lookup(self, cli_val, file_conf_val, default=None):
    if cli_val is not None:
        return cli_val
    if file_conf_val is not None:
        return file_conf_val
    return default
```

Here's the pseudo code.

An alternative solution is to have the `_lookup` receive only the key in the configuration file so that it's called like this:
`self.slack_token = self._lookup(slack_token, 'slack.token')`.

kopackiw:
This PR fixes the issue with ignoring "falsy" values when resolving attribution values hierarchy.

`_first_not_none` is a helper which will return first value which is not `None` (if it exists - otherwise, it falls back to `None`).

I went with `*args` to be able to pass all values that need to be checked as next arguments to the function, without any nesting. Thus, we can maintain hierarchy just be looking at the order of arguments.

Also, when `dict.get` cannot find value, it falls back to `None`. I've extracted the default value from this to outer scope, to achieve better readability (as mentioned above).

closes #336

### DIFF
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -46,39 +46,64 @@ class Config:
 
         config = self._load_configuration()
 
-        self.target_dir = config.get("target-path") or os.getcwd()
-
-        self.update_bucket_website = update_bucket_website or config.get(
-            "update_bucket_website", False
+        self.target_dir = _first_not_none(
+            config.get("target-path"),
+            os.getcwd(),
         )
-        self.timezone = timezone or config.get("timezone")
 
-        self.slack_webhook = slack_webhook or config.get(self._SLACK, {}).get(
-            "notification_webhook"
+        self.update_bucket_website = _first_not_none(
+            update_bucket_website,
+            config.get("update_bucket_website"),
+            False,
         )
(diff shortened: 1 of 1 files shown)


kopackiw:
I don't know why, but the `CI_PROFILES_YML` seems to be empty, thus failing whole pipeline.

elongl:
Great solution! Thank you very much for solving this problem. Merging.

PR #396, opened 2022-11-11
Let boto3 determine AWS credentials

hengpor:
Let boto3 determine AWS credentials. user only need to provide s3 bucket name if they want to send report to s3 and make other aws arguments optional

### DIFF
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -128,12 +128,6 @@ class Config:
         return self.slack_webhook or (self.slack_token and self.slack_channel_name)
 
     @property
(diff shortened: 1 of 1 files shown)


elongl:
Hi @hengpor ! Thanks for the contribution.
Could you please also delete the `has_aws` property as it's not used now?
Thanks.

elongl:
Hi @hengpor ! Thanks a lot for your contribution.
Could you please delete the `has_aws` property from `config.py` as it's no longer used anywhere.
I'll merge it right after 😄

PR #399, opened 2022-11-15
Introduce Elementary extension for Meltano into docs

SBurwash:


docs/deployment-and-configuration/elementary-in-production.mdx
```diff
@@ -67,6 +67,7 @@ Some options include:
 2. Airflow
 3. Prefect
 5. Using cron
+6. Using Meltano (DataOps ELT Platform) - Learn all about Meltano [here](https://meltano.com/) and start leveraging the full power of your DBT models with their [custom Elementary integration](https://hub.meltano.com/utilities/elementary).
```
Maayan-s, line 70: Thanks @SBurwash !
Few comments:
- There is a numbering problem (4 is missing). In terms of order, I think Meltano should be above cron (should probably be the last resort). 
- I don't want it to look like this is a preferable option or something, all are valid. Also, I assume that this is only for users that are already using Meltano, so they know what it is. So I think a better wording would be similar to what we wrote on the Github action, like: 
```
4. Meltano - Checkout the [custom Elementary integration](https://hub.meltano.com/utilities/elementary) developed by < see next comment>.
```
- I would also love to credit you for it, but I don't know if you want that, so feel free to complete with whatever feels right (could also be by the community if you don't want it to mention you). 

WDYT?

Maayan-s (approved):
Perfect! Merging!

PR #401, opened 2022-11-15
support test description both on the report and the alerts

IDoneShaveIt:
1. Extract meta field when query tests run results
2. Use the meta field to extract test description
3. Pass test description both to the UI and the alerts


elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
```diff
@@ -252,6 +253,7 @@
         test_results.test_params,
         test_results.severity,
         test_results.status,
+        tests.meta
```
IDoneShaveIt, line 256:
```suggestion
        tests.meta,
```

haritamar (approved):

PR #411, opened 2022-11-20
--dbt-quoting CLI flag

haritamar:


elementary/monitor/dbt_project/dbt_project.yml
```diff
@@ -29,6 +29,11 @@ clean-targets: # directories to be removed by `dbt clean`
 vars:
   edr_cli_run: true
 
+quoting:
+  database: "{{ env_var('DATABASE_QUOTING', 'None') | as_native }}"
+  schema: "{{ env_var('SCHEMA_QUOTING', 'None')  | as_native }}"
+  identifier: "{{ env_var('IDENTIFIER_QUOTING', 'None') | as_native }}"
```
haritamar, line 35: `as_native` is the only way I could think of to be able to pass either True, False or None as values

elementary/clients/dbt/dbt_runner.py
```diff
@@ -156,3 +160,15 @@ def test(
             command_args=command_args, vars=vars, quiet=quiet
         )
         return success
+
+    def _get_command_env(self):
+        env = os.environ.copy()
+        if self.dbt_quoting is not None:
```
IDoneShaveIt, line 166: I really like the option to pass env vars to the dbt runner!
IMO we should make it more generic and instead of passing a specific key that we need to manually map to the envs (and create a new cli variable for it), we should do something like that:
```
if self.dbt_env_vars is not None:
    env.update({**self.dbt_env_vars})
```

That way we can easily make new env vars supported in the future without changing a thing!
WDYT?

elementary/clients/dbt/dbt_runner.py
```diff
@@ -18,11 +19,13 @@ def __init__(
         profiles_dir: str,
         target: Optional[str] = None,
         raise_on_failure: bool = True,
+        dbt_quoting: bool = None,
```
IDoneShaveIt, line 22: See my comment on `_get_command_env`
I think it will be better to pass dbt_env_vars/env_vars

elementary/config/config.py
```diff
@@ -22,11 +25,22 @@ class Config:
...
+        },
+        "additionalProperties": False,
+    }
+
     def __init__(
         self,
         config_dir: str = DEFAULT_CONFIG_DIR,
         profiles_dir: str = DEFAULT_PROFILES_DIR,
         profile_target: str = None,
+        dbt_quoting: bool = None,
```
IDoneShaveIt, line 43: See my prev comments about making it a dict of env vars 🙂

elementary/monitor/cli.py
```diff
@@ -15,6 +15,15 @@
 
 # Displayed in reverse order in --help.
 def common_options(func):
+    func = click.option(
+        "--dbt-quoting",
```
IDoneShaveIt, line 19: In continue to my comments, we could except here for --env-vars which will be a string of a json/yaml of the env vars which we will load as dict and pass it to the command runner :)
We could also support a .env file path that will be loaded as the env vars. WDYT?

IDoneShaveIt (commented):
The PR looks good!
I have a different idea of more generic support of env vars.
Let me know what you think about it 🙂

elementary/monitor/cli.py
```diff
@@ -15,6 +15,15 @@
...
+        "-q",
```
elongl, line 20: Maybe worth saving `-q` for something more common than this?

elementary/clients/dbt/dbt_runner.py
haritamar: It's a cool idea!
My only thought is that in this case I'm passing internal env variables that I made up (e.g. not something that is directly supported by dbt)
So for this particular use case maybe an explicit CLI parameter is good? WDYT?

But nonetheless maybe I can change the code to support both.

elementary/monitor/cli.py
haritamar: Added a generic param in addition to the quoting one, as discussed on Slack

IDoneShaveIt (approved):
LGTM!

PR #413, opened 2022-11-21
Read result rows

elongl:
This is currently a breaking change as `result_rows` is queried from `elementary_test_results` which might not exist yet.
This can be bypassed by reading the Elementary's dbt package version and rendering the `result_rows` only if it's above a certain version.

### DIFF
--- a/elementary/clients/dbt/dbt_runner.py
+++ b/elementary/clients/dbt/dbt_runner.py
@@ -44,10 +44,11 @@ class DbtRunner:
         if vars:
             json_vars = json.dumps(vars)
             dbt_command.extend(["--vars", json_vars])
+        log_msg = f"Running {' '.join(dbt_command)}"
         if not quiet:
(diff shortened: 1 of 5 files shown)


elementary/clients/dbt/dbt_runner.py
```diff
@@ -47,10 +47,11 @@ def _run_command(
         if vars:
             json_vars = json.dumps(vars)
             dbt_command.extend(["--vars", json_vars])
+        log_msg = f"Running {' '.join(dbt_command)}"
         if not quiet:
```
haritamar, line 51: Not really related to the PR - but I'm not a fan of this "quiet" flag (I'd prefer that we'll change the logging level of the console handler based on a flag)

haritamar: hmm I see it's controlled per command and not by an outside flag, so maybe it's fine

haritamar (approved):

Issue #414, opened 2022-11-21
Search profiles.yml in the local directory before ~/.dbt


haritamar:
**Is your feature request related to a problem? Please describe.**
The default behavior of looking up for `profiles.yml` is to first look it up in the local directory and then to default to ~/.dbt/profiles.yml.
That being said, currently the EDR cli only considers the latter.

**Describe the solution you'd like**
Align the behavior with dbt by checking the local directory first.

PR #438, opened 2022-11-29
Letting dbt lookup the profiles.yml unless --profiles-dir specified.

elongl:
Fixes https://github.com/elementary-data/elementary/issues/414

### DIFF
--- a/elementary/clients/dbt/dbt_runner.py
+++ b/elementary/clients/dbt/dbt_runner.py
@@ -168,3 +168,7 @@ class DbtRunner:
         if self.dbt_env_vars is not None:
             env.update(self.dbt_env_vars)
         return env
+
+    def debug(self, quiet: bool = False) -> bool:
+        success, _ = self._run_command(command_args=["debug"], quiet=quiet)
+        return success
--- a/elementary/clients/dbt/dbt_runner.py
+++ b/elementary/clients/dbt/dbt_runner.py
@@ -16,10 +16,10 @@ class DbtRunner:
     def __init__(
         self,
         project_dir: str,
-        profiles_dir: str = None,
-        target: str = None,
+        profiles_dir: Optional[str] = None,
+        target: Optional[str] = None,
         raise_on_failure: bool = True,
-        dbt_env_vars: Dict[str, str] = None,
+        dbt_env_vars: Optional[Dict[str, str]] = None,
     ) -> None:
         self.project_dir = project_dir
         self.profiles_dir = profiles_dir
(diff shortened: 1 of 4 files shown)


elementary/config/config.py
```diff
@@ -166,31 +162,31 @@ def has_gcs(self):
...
 
-    def _validate_elementary_profile(self):
-        profiles_path = os.path.join(self.profiles_dir, "profiles.yml")
-        try:
-            profiles_yml = OrderedYaml().load(profiles_path)
-            if "elementary" not in profiles_yml:
-                raise NoElementaryProfileError
-        except FileNotFoundError:
-            raise NoProfilesFileError(self.profiles_dir)
+    def _validate_internal_dbt_project(self):
```
elongl, line 182: This is also not entirely necessary.
If we remove it then something else will raise `DbtCommandError`.
The problem is that on certain things we ignore the error and continue.
This is safer in the sense that if it fails, everything else is _guaranteed_ to fail as well.

elementary/clients/dbt/dbt_runner.py
```diff
@@ -16,8 +16,8 @@ class DbtRunner:
     def __init__(
         self,
         project_dir: str,
-        profiles_dir: str,
-        target: Optional[str] = None,
+        profiles_dir: str = None,
+        target: str = None,
```
IDoneShaveIt, line 20: Well those are optional vars, so why remove the Optional?
It is really helpful to have a good typing!

elementary/config/config.py
```diff
@@ -166,31 +162,31 @@ def has_gcs(self):
...
+        self._validate_internal_dbt_project()
         self._validate_timezone()
         if not self.has_slack:
             raise InvalidArgumentsError(
                 "Either a Slack token and a channel or a Slack webhook is required."
             )
 
     def validate_report(self):
-        self._validate_elementary_profile()
+        self._validate_internal_dbt_project()
```
IDoneShaveIt, line 173: Maybe instead of running it for each method (and remember to add it to a new method), we will run it on the init?
That way no matter what, before we start anything we will validate the internal project?

elementary/config/config.py
IDoneShaveIt: Sounds right!

IDoneShaveIt (commented):
Looking good!
Left some small comments 🙂

elementary/config/config.py
elongl: I thought so as well.
The problem is that we also use it [here](https://github.com/elementary-data/elementary/blob/e52e384ad264c43ebcab7221edf2eeec64374526/elementary/cli/cli.py#L46) and we don't want to make that validation there.

elongl: We could also make it validate by default and only opt-out there.

Issue #434, opened 2022-11-28
fix an issue when auto open report in WSL


ivan-toriya:
[Slack Thread](https://elementary-community.slack.com/archives/C02CTC89LAX/p1669034786609089?thread_ts=1669034786.609089&cid=C02CTC89LAX)

"Hi, im new to using Elementary, i've set up all the steps and when i try to run: edr monitor report i get this:

The error msg is a little bit strange since it says it cannot find the file specified. But elementary has generated a elementary.html file which is shown in the screenshot."

![image](https://user-images.githubusercontent.com/43750521/204319272-803cdead-7411-42fd-b738-7dc757b2c492.png)

ivan-toriya:
@elongl suggested
"we attempt to open it and if it fails to log it and continue so that it doesn't raise an error.
This is what [dbt does](https://github.com/dbt-labs/dbt-core/blob/43e3fc22c4eae4d3d901faba05e33c40f1f1dc5a/core/dbt/task/serve.py#L34) as well.
...
So basically no need to deal with WSL specifically.
...
The code that needs to be changed is [here](https://github.com/elementary-data/elementary/blob/ac8e4f92bda34ada1a2e168502b96acd9f8441ea/elementary/monitor/data_monitoring.py#L242).
Basically I think what we should do is wrap it in a try..except webrowser.Error and log if an occurs."

elongl:
Thanks a lot for opening this issue.
Looking forward to the pull request.

elongl:
Solved in https://github.com/elementary-data/elementary/pull/447.

PR #447, opened 2022-11-30
fix an issue when auto open report in WSL

ivan-toriya:
It solves the https://github.com/elementary-data/elementary/issues/434

### DIFF
--- a/elementary/monitor/data_monitoring.py
+++ b/elementary/monitor/data_monitoring.py
@@ -235,7 +235,11 @@ class DataMonitoring:
             elementary_output_json_file.write(dumped_output_data)
 
         if should_open_browser:
-            webbrowser.open_new_tab("file://" + html_path)
+            try:
+                webbrowser.open_new_tab("file://" + html_path)
+            except webbrowser.Error as e:
+                logger.exception(f"Could not open {html_path} in the browser - Error: {e}")
+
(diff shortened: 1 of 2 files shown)


elementary/monitor/data_monitoring.py
```diff
@@ -235,7 +235,11 @@ def generate_report(
             elementary_output_json_file.write(dumped_output_data)
 
         if should_open_browser:
-            webbrowser.open_new_tab("file://" + html_path)
+            try:
+                webbrowser.open_new_tab("file://" + html_path)
+            except webbrowser.Error as e:
+                logger.exception(f"Could not open {html_path} in the browser - Error: {e}")
```
elongl, line 241: I think we can skip logging the exception itself using `logger.exception` as it produces very large, and a bit too detailed, tracebacks which are not interesting in this case.
This error can occur only on rare occasions which are mostly when you use WSL.

What do you think?

```suggestion
                logger.error("Unable to open the web browser.")
```

elongl (commented):
Thanks a lot Ivan!
This is really useful with the growing usage of WSL.
Left one small comment and I'll merge right after.

elementary/monitor/data_monitoring.py
ivan-toriya: Agree

PR #448, opened 2022-11-30
Add email parsing for slack handle alerts

jelstongreen:
This is needed for Slack Enterprise customers for whom simple `@josh.elston-green` tags do not work and need to use the Slack User IDs like `<@U123456>`

### DIFF
--- a/elementary/clients/slack/client.py
+++ b/elementary/clients/slack/client.py
@@ -60,6 +60,12 @@ class SlackClient(ABC):
         raise NotImplementedError
 
 
+    @abstractmethod
+    def get_user_id_from_email(self, email: str):
+        raise NotImplementedError
+
+
+
 class SlackWebClient(SlackClient):
     def _initial_client(self):
         return WebClient(token=self.token)
@@ -111,6 +117,14 @@ class SlackWebClient(SlackClient):
         else:
             logger.error("Failed to send report to Slack.")
         return send_succeed
+
+    def get_user_id_from_email(self, email: str) -> str:
+        logger.info(f"Attempting to get slack id of user: {email}")
+        try:
+            return self.client.users_lookupByEmail(email=email)["user"]["id"]
+        except SlackApiError as e:
+            logger.error(f"Slack client error: {e.response['error']}")
+        return ""
 
     def _get_channel_id(self, channel_name: str) -> Optional[str]:
         try:
@@ -206,4 +220,7 @@ class SlackWebhookClient(SlackClient):
         raise NotImplementedError
 
     def send_report(self, **kwargs):
+        raise NotImplementedError
+
+    def get_user_id_from_email(self, **kwargs):
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


jelstongreen:
So for this schema.yml:
image

Here are the alerts:
image

This shows the two handles coming through in the correct format with one working which is the UID and the other not working as we use Slack Enterprise 

The model failure as no owners which I think is a seperate bug out of scope.

elementary/clients/slack/client.py
```diff
@@ -112,6 +118,14 @@ def send_report(self, channel: str, report_file_path: str):
             logger.error("Failed to send report to Slack.")
         return send_succeed
 
+    def get_user_id_from_email(self, email: str) -> str:
```
haritamar, line 121: Can you add a cache to this function? e.g. something like:
```
if email not in self.email_to_user_id_cache:
     self.email_to_user_id_cache[email] = self.client.users_lookupByEmail(email=email)["user"]["id"] 
return self.email_to_user_id_cache[email]
```

Mainly want to avoid having to wait several times for slack API calls on the same user for different alerts.

setup.py
```diff
@@ -38,6 +38,8 @@
         "pydantic<2.0",
         "networkx>=2.3,<3",
         "packaging>=20.9,<22.0",
+        "regex"
+        ""
```
haritamar, line 42: Remove the empty string?

elementary/clients/slack/client.py
```diff
@@ -112,6 +118,14 @@ def send_report(self, channel: str, report_file_path: str):
...
+        logger.info(f"Attempting to get slack id of user: {email}")
+        try:
+            return self.client.users_lookupByEmail(email=email)["user"]["id"]
```
haritamar, line 124: What happens when the email is not found? Is the response None or will it cause a `SlackApiError`?

elementary/monitor/data_monitoring.py
```diff
@@ -83,13 +84,32 @@ def __init__(
...
+
+        def _regex_match_owner_email(potential_email_str):
+            email_regex = r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'
+
+            return re.fullmatch(email_regex, potential_email_str)
+
+        if owners_str != [] and owners_str != '':
+            owners_list = [owner.strip() for owner in owners_str.split(',')]
+            owners_validated = [
+                f'<@{self.slack_client.get_user_id_from_email(owner)}>' if _regex_match_owner_email(owner) else owner
```
haritamar, line 97: You should check if the response from `get_user_id_from_email` is empty first, and in that case keep the owner as is (right now an invalid mail will cause you to save `<@>`)

elementary/clients/slack/client.py
jelstongreen: It returns None unless there is an error with the call to the SlackAPI - a missing email just returns the original handle with the latest changes.

PR #452, opened 2022-12-04
tags and owners sidebars

IDoneShaveIt:
Added 2 more sidebars to the report - tags & owners.


elementary/monitor/api/sidebar/sidebar.py
```diff
@@ -48,3 +68,33 @@ def _update_dbt_sidebar(
                 if part not in dbt_sidebar:
                     dbt_sidebar[part] = {}
                 dbt_sidebar = dbt_sidebar[part]
+
+    def get_tags_sidebar(
+        self,
+        models: Dict[str, NormalizedModelSchema],
+        sources: Dict[str, NormalizedModelSchema],
+    ) -> TagsSidebarSchema:
```
haritamar, line 76: Maybe we can somehow merge these two functions?
E.g. `get_sidebar_by_labels` or something

(Not very critical, as these are very short functions)

IDoneShaveIt: I like the seperation because this is an equivalent to an API.
I can make a common method that is used by both.

haritamar: Not critical, just my automatic duplication radar :)
I think it's also fine as-is as it's very clear

IDoneShaveIt: That is a good radar!

elementary/monitor/api/tests/schema.py
```diff
@@ -53,11 +83,8 @@ def format_time_utc(cls, time_utc):
         return convert_partial_iso_format_to_full_iso_format(time_utc)
 
 
-class TotalsInvocationsSchema(BaseModel):
-    errors: Optional[int] = 0
-    warnings: Optional[int] = 0
-    passed: Optional[int] = 0
-    resolved: Optional[int] = 0
+class TotalsInvocationsSchema(TotalsSchema):
+    pass
```
haritamar, line 87: Why do we need both `TotalsInvocationsSchema` and `TotalsSchema`?

IDoneShaveIt: Well they used to be different and I somehow like that we define a schema for each data type (even if they are the same).
But I agree that currently its a bit redundant

elementary/monitor/api/filters/filters.py
```diff
@@ -0,0 +1,103 @@
...
+                errors_filter.add_model_unique_id(model_unique_id)
+            if total.passed:
+                passed_filter.add_model_unique_id(model_unique_id)
+            if (
+                not total.failures
+                and not total.warnings
+                and not total.errors
+                and not total.passed
+            ):
+                no_tests_filter.add_model_unique_id(model_unique_id)
```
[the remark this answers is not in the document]
IDoneShaveIt: Yes :) 
You can define models without tests / run dbt test -s which will cause only partial tests to run.
On both cases we can have a situation where there are no tests 🙂

PR #457, opened 2022-12-06
Fix tracking

elongl:


elementary/tracking/anonymous_tracking.py
```diff
@@ -53,17 +54,9 @@ def init(self):
             pass
 
     def _get_anonymous_user_id(self):
-        legacy_user_id_path = Path().joinpath(
-            self.config.profiles_dir, self.ANONYMOUS_USER_ID_FILE
```
elongl, line 57: This is `None` by default which raises an Exception.

haritamar (approved):

PR #459, opened 2022-12-07
Still get user ids when only a non-list is passed to owner/subscriber

jelstongreen:


elementary/monitor/data_monitoring.py
```diff
@@ -100,7 +101,15 @@ def _get_user_id(email: str) -> str:
...
 
-        if isinstance(emails, list) and emails != []:
+        def _parse_str_to_list(emails: str) -> List[str]:
+            try:
+                return ast.literal_eval(emails)
+            except:
+                return list(emails.split(","))
+
+        if isinstance(emails, str):
+            emails = _parse_str_to_list(emails)
```
IDoneShaveIt, line 111: I think that we can simplifier the code a bit in the following way:
Instead of doing the if statement like that, you can do `emails = _parse_str_to_list(emails) if isinstance(emails, str) else emails`.
That way emails will always be a list and you could skip the next if statement and always run:
```python
ids = [
    _get_user_id(email) if _regex_match_owner_email(email) else email
    for email in emails
]
parsed_ids_str = prettify_json_str_set(ids)
return parsed_ids_str
```

IDoneShaveIt: The finale code will look like:
```python
def _parse_str_to_list(emails: str) -> List[str]:
    try:
        return ast.literal_eval(emails)
    except:
        return list(emails.split(","))

emails = _parse_str_to_list(emails) if isinstance(emails, str) else emails
ids = [
    _get_user_id(email) if _regex_match_owner_email(email) else email
    for email in emails
]
parsed_ids_str = prettify_json_str_set(ids)
return parsed_ids_str
```

IDoneShaveIt (commented):
Thanks for raising this issue and for the contribution!
The solution looks great, just added some small comments 🙂

elementary/monitor/data_monitoring.py
```diff
@@ -91,7 +92,7 @@ def __init__(
         self.send_test_message_on_success = send_test_message_on_success
         self.disable_samples = disable_samples
 
-    def _parse_emails_to_ids(self, emails: List[str]) -> str:
+    def _parse_emails_to_ids(self, emails: Union[str, List[str]]) -> str:
```
elongl, line 95: Shouldn't that logic be in `SlackClient` or `AlertsAPI` rather than `DataMonitoring`?

PR #464, opened 2022-12-10
Created an --env flag.

elongl:


elementary/monitor/cli.py
```diff
@@ -46,6 +46,12 @@ def common_options(func):
         default=7,
         help="Set a limit to how far back should edr collect data.",
     )(func)
+    func = click.option(
+        "--env",
+        type=str,
+        default="dev",
+        help="The environment that you're running against. Usually this would be either 'dev' or 'prod'.",
```
oravi, line 53: cosmetics - This flag indicates if you are running Elementary in prod or dev environment and will be reflected accordingly in the report. Value should be either 'dev' or 'prod' (default is 'dev').

elementary/monitor/cli.py
```diff
@@ -46,6 +46,12 @@ def common_options(func):
         default=7,
         help="Set a limit to how far back should edr collect data.",
     )(func)
+    func = click.option(
+        "--env",
```
oravi, line 50: Can you use click to enforce only 'dev' or 'prod' for now?

PR #478, opened 2022-12-19
Using project directory to run the artifacts.

elongl:


elementary/monitor/data_monitoring.py
```diff
@@ -514,8 +532,11 @@ def _check_dbt_package_compatibility(dbt_pkg_ver: str):
...
             or dbt_pkg_ver.minor != py_pkg_ver.minor
         ):
-            click.secho(
+            logger.warning(
                 f"You are using incompatible versions between edr ({py_pkg_ver}) and Elementary's dbt package ({dbt_pkg_ver}).\n "
                 "Please upgrade the major and minor versions to align.\n",
-                fg="yellow",
             )
+
+    def run_artifacts(self):
```
elongl, line 540: Should run all of Elementary's models?
Will take longer but solves compatibility issues in case we added columns etc.

IDoneShaveIt: I think so.
Our models are relative small and will add small amount of running time IMO.
So lets make sure we are fully updated :)

elementary/monitor/data_monitoring.py
```diff
@@ -68,18 +67,34 @@ def __init__(
...
+        if self.config.project_dir:
+            self.user_dbt_runner = DbtRunner(
+                self.config.project_dir,
+                self.config.profiles_dir,
+                self.config.profile_target,
+                dbt_env_vars=self.config.dbt_env_vars,
+            )
+        else:
+            self.user_dbt_runner = None
+            logger.warning(
```
IDoneShaveIt, line 85: Haven't we talked about cloud users that we don't want to tell the that they must provide a local dir yet?

IDoneShaveIt (commented):
Looking good!
Small comments

PR #489, opened 2022-12-26
fix alerts macros redshift bug

IDoneShaveIt:


elementary/monitor/dbt_project/macros/alerts/get_model_alerts.sql
```diff
@@ -27,37 +27,43 @@
...
+                alerts_in_time_limit.status,
+                alerts_in_time_limit.full_refresh,
+                {# backwards compatibility #}
+                case
+                    when alerts_in_time_limit.suppression_status is NULL and alerts_in_time_limit.alert_sent = TRUE then 'sent'
+                    when alerts_in_time_limit.suppression_status is NULL and alerts_in_time_limit.alert_sent = FALSE then 'pending'
+                    else suppression_status
+                end as suppression_status,
+                case 
+                    when alerts_in_time_limit.sent_at is NULL then '1970-01-01 00:00:00'
```
elongl, line 55: Why not keep it at `NULL`?
I don't see any reason we'd use the UNIX timestamp here.

elementary/monitor/dbt_project/models/alerts/alerts.sql
```diff
@@ -33,7 +33,7 @@ select
     *,
     false as alert_sent,  {# backwards compatibility #}
```
elongl, line 34: I don't think this is needed.
If we'd remove it, it'll default to `NULL` which is okay.

(Same for all alert tables)

IDoneShaveIt: The problem is that because we have backwards compatibility, we check the content of the column so we could set the relevant suppression status for old alerts.
So if we'll remove the column, new users won't have it in their tables and we will get an SQL error for trying to read an not existing column.

So if we need to keep this column for backwards compatibility, lets keep the alert_sent with relevant data instead of nulls 🙂

elementary/monitor/dbt_project/macros/alerts/get_model_alerts.sql
IDoneShaveIt: Yeah You are right, this is a leftover from when I developed the feature and didn't handle yet NULL values 🙂

PR #492, opened 2022-12-27
Fix duplicate rows bug

IDoneShaveIt:
Handle tests with dynamic sub_type + column bug - show duplicates rows for the same test on the UI test screens


elementary/monitor/api/tests/tests.py
```diff
@@ -41,16 +41,6 @@ def __init__(self, dbt_runner: DbtRunner):
         super().__init__(dbt_runner)
         self.invocations_api = InvocationsAPI(dbt_runner)
 
-    @staticmethod
-    def get_test_sub_type_unique_id(
```
IDoneShaveIt, line 45: Moved into current_tests_run_results_query macro to simplify it and having one place that generates the `test_sub_type_unique_id`

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
```diff
@@ -245,6 +257,16 @@
     
     select
         test_results.id,
+        {#
```
IDoneShaveIt, line 260: Moved `test_sub_type_unique_id` into the query and added support for the use cases of `schema changes` and `dimension anomalies`

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
```diff
@@ -186,6 +188,16 @@
 
     select
         test_results.id,
+        {#
```
IDoneShaveIt, line 191: Moved `test_sub_type_unique_id` into the query and added support for the use cases of `schema changes` and `dimension anomalies`

elementary/monitor/api/tests/tests.py
```diff
@@ -202,7 +195,15 @@ def get_test_results(
             test_results_metadata = self.get_tests_metadata(days_back=days_back)
 
         tests_sample_data = self.get_run_cache(TESTS_SAMPLE_DATA)
-        if tests_sample_data is None:
+        if invocation.invocation_id:
```
IDoneShaveIt, line 198: Get only tests samples for the filtered invocation and not all the latest test runs (improve report performances)

elementary/monitor/dbt_project/macros/get_test_results.sql
```diff
@@ -1,24 +1,46 @@
...
+            )
+            where row_number = 1
         ),
 
         latest_tests_in_the_last_chosen_days as (
-            select * from tests_in_last_chosen_days where row_number = 1
+            select 
+                tests.*,
+                {{ elementary.datediff(elementary.cast_as_timestamp('tests.detected_at'), elementary.current_timestamp(), 'day') }} as days_diff
+            from test_results tests
```
oravi, line 29: why do you need an alias here?

elementary/monitor/dbt_project/macros/get_test_results.sql
```diff
@@ -1,24 +1,46 @@
...
-        tests_in_last_chosen_days as (
-            select *,
-                  {{ elementary.datediff(elementary.cast_as_timestamp('detected_at'), elementary.current_timestamp(), 'day') }} as days_diff,
-                  row_number() over (partition by model_unique_id, test_unique_id, column_name, test_sub_type order by detected_at desc) as row_number
+        test_latest_invocations as (
+            select 
+                test_unique_id,
+                invocation_id
+            from (
+                select
```
oravi, line 16: Why is it a sub query and not a CTE?

tests/integration/report/test_report.py
```diff
@@ -62,6 +62,297 @@ def test_sidebar(report_data_fixture):
...
 
+def test_duplicate_rows_for_latest_run_status(report_data_fixture):
+    report_data = get_report_data()
+    # e2e contains 2 failed dimension tests and 1 passed
+    assert_test_counter(
+        report_data=report_data,
+        test_type="anomaly_detection",
+        name="dimension_anomalies",
+        status="fail",
+        expected_ammount=2,
```
oravi, line 73: Let's change the counter to be more around validating the test sub types (for each sub type we should validate exactly one row)

elementary/monitor/api/tests/tests.py
```diff
@@ -149,18 +140,19 @@ def get_test_info_from_test_metadata(
...
-                test_params=try_load_json(metadata.test_params),
+                test_params=metadata.test_params,
             )
         else:
             configuration = dict(
                 test_name=metadata.test_name,
-                timestamp_column=test_params.get("timestamp_column"),
-                testing_timeframe=test_params.get("timeframe"),
-                anomaly_threshold=test_params.get("sensitivity"),
+                timestamp_column=metadata.test_params.get("timestamp_column"),
```
oravi, line 148: Are you sure metadata.test_params exists / not null?

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
```diff
@@ -26,248 +12,57 @@
...
+        elementary_test_results.id,
+        {#
+            Due to inconsistency test sub type and column name in some of elementary's tests
+            we need to calculate different test sub type unique id which is used to identify between different tests.
+        #}
+        case
+            when elementary_test_results.test_type = 'schema_change' then elementary_test_results.test_unique_id
+            {# In old versions of elementary, elementary_test_results doesn't contain test_short_name, so we use dbt_test short_name. #}
+            when dbt_tests.short_name = 'dimension_anomalies' then elementary_test_results.test_unique_id
+            else coalesce(elementary_test_results.test_unique_id, 'None') || '.' || coalesce(elementary_test_results.test_sub_type, 'None')
```
oravi, line 33: if it's a dbt test we should use the test_unique_id, no?

elementary/monitor/api/tests/tests.py
IDoneShaveIt: Yeah because by default we set it to `{}` if its empty (on the pydantic schema)

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
IDoneShaveIt: The subtype of dbt test is always the same + each test is mapped to exactly one elementary_test_results, so it is fine to use the `test_unique_id` + `test_sub_type` for tests that aren't ours.
One key generation logic to rule them all!
(It is simpler and better IMO)

PR #498, opened 2022-12-29
Alerts filters

IDoneShaveIt:
Support filtering alerts using --select argument.
In case project dir is provided, we use dbt ls command to filter the nodes.
Otherwise we support manual filtering over tag / owner / model


elementary/monitor/cli.py
```diff
@@ -123,6 +123,22 @@ def decorator(func):
...
         )(func)
+        func = click.option(
+            "--project-profile-target",
+            type=str,
+            default=None,
+            help="Which target to load for the given profile. "
+            "If specified, the target will be used for your dbt project."
+            "Else, the --profile-target will be used.",
+        )(func)
+        func = click.option(
```
haritamar, line 134: I don't think it's critical for now but I wonder if people will expect multiple selectors to work as well.
Can be added to a future version if needed though.

elementary/monitor/data_monitoring/data_monitoring_alerts.py
```diff
@@ -30,17 +32,64 @@ def __init__(
...
+            self.elementary_database_and_schema,
         )
         self.sent_alert_count = 0
         self.send_test_message_on_success = send_test_message_on_success
 
+    def _parse_filter(
+        self, filter: Optional[str] = None
+    ) -> Optional[DataMonitoringAlertsFilter]:
+        if filter:
+            if self.user_dbt_runner:
```
haritamar, line 56: You still need to handle the owner selector in this case as well, no?
As it's not a supported dbt one

elementary/monitor/data_monitoring/data_monitoring_alerts.py
```diff
@@ -30,17 +32,64 @@ def __init__(
...
+                self.tracking.set_env("select_method", "dbt selector")
+                selector_api = SelectorAPI(self.user_dbt_runner)
+                node_names = selector_api.get_selector_results(selector=filter)
+                return DataMonitoringAlertsFilter(node_names=node_names)
+
+            else:
+                data_monitoring_filter = DataMonitoringAlertsFilter()
+                tag_regex = re.compile(r"tag:.*")
+                owner_regex = re.compile(r"owner:.*")
+                model_regex = re.compile(r"model:.*")
```
haritamar, line 66: Mentioning here as well - when selecting models in dbt you don't prefix with `model:`
So it's inconsistent with the dbt project mode (and how people are used to selectors)

So I think it would also make sense to support `edr monitor -s customers` rather than `edr monitor -s model:customers`

elementary/clients/dbt/dbt_runner.py
```diff
@@ -172,3 +173,29 @@ def _get_command_env(self):
...
+                and try_load_json(command_outputs[0])
+                and try_load_json(command_outputs[1])
+            ):
+                logger.warning(
+                    f"The selection criterion '{select}' does not match any nodes"
+                )
+                return []
+            # When nodes are matched, ls command returns strings of the node names.
+            else:
+                return command_outputs
```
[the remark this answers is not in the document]
IDoneShaveIt: If nodes are matched, the output of dbt will be:
"node1\nnode2\n"
So I don't see why we should return it like that as well.

When there are no nodes the output is always the same and our warning explains the same thing so it is redundant and will never be used IMO.

WDYT?

elementary/monitor/cli.py
IDoneShaveIt: I think that dbt supports only a single selector (when passing more then one it uses the latest one), so I think that we should see the usage of it and wait for users' requests 🙂

elementary/monitor/data_monitoring/data_monitoring_alerts.py
IDoneShaveIt: @elongl showed me a very cool trick that works for dbt selector:
`dbt ls -s config.meta.owner:idan`

We decided that for now we will only document the non dbt project path, and if a user will ask for advance selector we will refer him to the dbt selector and will explain how to use it 🙂

elementary/clients/dbt/dbt_runner.py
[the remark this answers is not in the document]
haritamar: I mean something else :)
It's good that you split the rows, but I don't get why you call `try_load_json` and then return the non-json loaded values - why not return the output of `try_load_json`?

elementary/monitor/cli.py
haritamar: They support multiple:
image

But again - I don't see it as critical for now.

IDoneShaveIt: Oh I was thinking about providing multiple tags / owners.
I think it could be step 2

elementary/monitor/data_monitoring/data_monitoring_alerts.py
haritamar: So that's cool, but I still don't like that the interfaces for dbt project and non-dbt project are not the same. It means that things will break if you move to the dbt project mode.

So in my mind either you:
1. Support `owner:` in both.
2. Change the interface of the second to be: `config.meta.owner:`

elementary/clients/dbt/dbt_runner.py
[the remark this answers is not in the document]
IDoneShaveIt: Because I only use `try_load_json` to understand that I have no nodes matches.
Those json has no additional information, and has no use.
I prefer keeping it all simple and the same typing wise:
If there are matches, return a list of the matches.
If there are no matches, log an error and return an empty list.

There is no value by returning 2 dicts that has message that said that there are no node matches and then parse it outside the function.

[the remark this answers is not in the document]
haritamar: Ah OK, I see what you mean
I misunderstood this. Sababa.

PR #515, opened 2023-01-03
Add endpoint url as option for sending report to S3

### LINKED ISSUE #507: [ELE-29] Being able to define "endpoint_url" when sending report to S3
We have selfhosted S3 instances (not AWS). My need is  to send the generated report to the selfhosted one, instead of AWS.

As I understand it, boto3 is used, and it is possible to use other endpoints than aws with the parameter "endpoint_url". Ideal solution (for me) would be to able to choose custom endpoint with edr cli

<sub>[ELE-29](https://linear.app/elementary/issue/ELE-29/being-able-to-define-endpoint-url-when-sending-report-to-s3)</sub>

esenilsson:
Attempt to fix #507 by 

- Change config
- Change CLI
- Change client

To cater for s3 endpoint url

### DIFF
--- a/elementary/cli/cli.py
+++ b/elementary/cli/cli.py
@@ -1,567 +1,69 @@
-import sys
+import os
+from typing import Any
 
 import click
+from pyfiglet import Figlet
 
+import elementary.cli.upgrade
 from elementary.config.config import Config
-from elementary.monitor.data_monitoring.data_monitoring_alerts import (
-    DataMonitoringAlerts,
-)
-from elementary.monitor.data_monitoring.data_monitoring_report import (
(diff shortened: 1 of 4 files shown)


elongl:
Hi @esenilsson !
This looks great.
I’ll test it out tomorrow and merge as soon as possible.
Thanks a lot for your contribution.

elongl:
Hi @esenilsson,
This works great, thanks a lot.
There's one small change that we need to pass `endpoint_url` to the `client` rather than the `session` but given that it's very minor I'll fix it on our end.
Appreciate the contribution, it'll be released on the next version.

PR #548, opened 2023-01-10
ELE-142 unite tests queries - Relays on duplicate rows fix

IDoneShaveIt:
Unite between the queries `get_test_results`, `get_test_invocations` and `get_test_sample_data`.
This preventing a concurrency bugs, and improve the report performance.


elementary/monitor/data_monitoring/data_monitoring_report.py
```diff
@@ -132,14 +131,14 @@ def generate_report(
...
 
-            self.execution_properties["elementary_test_count"] = len(
-                [
-                    test_metadata
-                    for test_metadata in tests_metadata
-                    if test_metadata.test_type != "dbt_test"
-                ]
-            )
-            self.execution_properties["test_result_count"] = len(tests_metadata)
+            # self.execution_properties["elementary_test_count"] = len(
```
elongl, line 134: Accident?

elementary/monitor/api/tests/schema.py
```diff
@@ -27,10 +27,13 @@ class DbtTestResultSchema(BaseModel):
     failed_rows_count: Optional[int] = None
 
 
-class TestMetadataSchema(BaseModel):
+class RawTestResultSchema(BaseModel):
```
elongl, line 30: Why is it called "Raw"?

elementary/monitor/api/tests/schema.py
```diff
@@ -50,6 +53,12 @@ class TestMetadataSchema(BaseModel):
     status: str
     test_created_at: Optional[str] = None
     days_diff: float
+    invocations_order: int
+    sample_data: Optional[Union[dict, List]] = None
+
+    # pydantic has a bug with Union fields. This is how to support it.
```
elongl, line 59: This is not a bug, that's intended behavior.
I think we should remove this comment.

elementary/monitor/api/tests/schema.py
```diff
@@ -50,6 +53,12 @@ class TestMetadataSchema(BaseModel):
     status: str
     test_created_at: Optional[str] = None
     days_diff: float
+    invocations_order: int
+    sample_data: Optional[Union[dict, List]] = None
```
elongl, line 57: Why can this be both a dict and a list?

elementary/monitor/api/tests/schema.py
```diff
@@ -50,6 +53,12 @@ class TestMetadataSchema(BaseModel):
     status: str
     test_created_at: Optional[str] = None
     days_diff: float
+    invocations_order: int
```
elongl, line 56: I don't understand what does this represent.

elementary/monitor/api/tests/schema.py
elongl, line 30: I think Pydantic models are not meant to have methods like that.
I think their main purposes should be data classes, in the sense that they hold validated data.

elementary/monitor/api/tests/tests.py
```diff
@@ -41,35 +41,32 @@ def __init__(self, dbt_runner: DbtRunner):
...
-        tests_metadata = (
+        test_results = (
             json.loads(run_operation_response[0]) if run_operation_response else []
         )
-        tests_metadata = [
-            TestMetadataSchema(**test_metadata) for test_metadata in tests_metadata
+        test_results = [
+            RawTestResultSchema(**test_result) for test_result in test_results
         ]
         if should_cache:
```
elongl, line 67: I personally think that we should delete this cache mechanism and create a coherent data flow in the program instead.

elementary/monitor/api/tests/tests.py
```diff
@@ -186,120 +165,118 @@ def get_test_results(
...
+                        status=raw_test_result.status,
                         affected_rows=self._parse_affected_row(
-                            results_description=test_invocation[
-                                "test_results_description"
-                            ]
+                            results_description=raw_test_result.test_results_description
                         ),
                     )
                 )
             except Exception:
```
elongl, line 265: Why do we have an `except Exception` here?

elementary/monitor/api/tests/tests.py
```diff
@@ -341,10 +318,10 @@ def _parse_affected_row(results_description: str) -> Optional[int]:
 
     def get_total_tests_results(
         self,
-        tests_info: Optional[List[TestInfoSchema]] = None,
+        tests_metadata: Optional[List[TestMetadataSchema]] = None,
```
elongl, line 321: Why is this `Optional`?

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
```diff
@@ -1,23 +1,21 @@
-{% macro current_tests_run_results_query(days_back = none, invocation_id = none) %}
+{% macro current_tests_run_results_query(days_back = none) %}
```
elongl, line 1: Why was this removed?
Isn't it necessary for the filters?

elongl (commented):
Could you share a bit of context about the high-level logic changes presented in this PR?
I'm troubling a bit to understand how we solved the duplicates issue.
Thanks a lot.

IDoneShaveIt:
@elongl hey!
This PR is basing on the duplicate rows fix PR.
The reason I didn't want people to review it yet, is because I didn't want to duplicate the thread about it 🙂
The changes here are only relevant for unifying the 3 queries over the elementary test reaults table, which fixes race conditions, missing keys bug and improve the performance of the report.

elementary/monitor/data_monitoring/data_monitoring_report.py
IDoneShaveIt: Yes!
Will return it!

elementary/monitor/api/tests/schema.py
IDoneShaveIt: So we have 2 different schemas for test result.
One is the rows of the table elementary test results, and one is for the actual test results object we have in the report.

Can change the naming into "TestResultRowSchema" or "DbTestResultSchema" 🙂

elementary/monitor/api/tests/schema.py
IDoneShaveIt: Good question!
Currently we have a different samlpe_data structure for different test types (The UI is based on that).
This is something that we have from day 1, and we need to unify, but this isn't part of this PR 🙂

elementary/monitor/api/tests/schema.py
IDoneShaveIt: I am not sure what do you mean by methods like that?
Maybe the comment is on the wrong line?

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
IDoneShaveIt: We do the filter now on the python level.
Instead of query the same data 3 times (1 for test results, 1 for invocations, 1 for data samples) we now query it only once, and filter the data when needed on the relevant method.

Invocation is not filtering the invocation part (test runs), so we need to do this filter outside of the query.

elementary/monitor/api/tests/schema.py
IDoneShaveIt: So this is part of the change in unifying the test queries.
Until now we had 3 different queries that were the same except the final filters (where clause).
We ordered the result of each test from the newest to oldest (invocations order), and we used that on the last filtering to get the latest tests runs + invocations for each test.

Now that we filter the results outside of the query (in the python), we need to return the invocations order from the query 🙂

elementary/monitor/api/tests/schema.py
IDoneShaveIt: I moved the more logical methods outside of the schema, so now it contains only data validation 🙂

elementary/monitor/api/tests/tests.py
IDoneShaveIt: Deleted!

elementary/monitor/api/tests/tests.py
IDoneShaveIt: This is here for a long time.
In case we couldn't parse an invocation, we didn't want to fail the whole report.
Now we are braver and changed our approach for breaking the report, but I am not sure if we want to remove that completely an fail on a single invocation 🤔 

My hunch is to remove it, but want to make sure first

elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
elongl: Awesome!

elementary/monitor/data_monitoring/data_monitoring_report.py
```diff
@@ -103,22 +103,23 @@ def generate_report(
             models = self.models_api.get_models(exclude_elementary_models)
             sources = self.models_api.get_sources()
             exposures = self.models_api.get_exposures()
+
             models_runs = self.models_api.get_models_runs(
                 days_back=days_back, exclude_elementary_models=exclude_elementary_models
             )
-            tests_metadata = self.tests_api.get_tests_metadata(days_back=days_back)
 
+            test_results_db_rows = self.tests_api.get_all_test_results_db_rows(
```
elongl, line 111: Just an idea, what do you think about having a `TestsAPI.get_tests(...)` that returns, roughly, an object that looks like this?

```py
class TestResults
    results: ...
    runs: ...
```

That way there's only a single call to `TestsAPI` to get test-related data and there's no need to re-normalize the data in `DataMonitoring`. Not very critical though.

elementary/monitor/data_monitoring/data_monitoring_report.py
```diff
@@ -103,22 +103,23 @@ def generate_report(
...
+            )
             (
                 test_results,
                 test_results_totals,
                 invocation,
-            ) = self._get_test_results_and_totals(
-                days_back=days_back,
-                disable_passed_test_metrics=disable_passed_test_metrics,
-            )
+            ) = self._get_test_results_and_totals(test_results_db_rows)
```
elongl, line 120: Can we maybe change those to return objects and not tuples?
(I know it was like that beforehand.)

elementary/monitor/data_monitoring/data_monitoring_report.py
```diff
@@ -238,21 +243,18 @@ def _get_lineage(self, exclude_elementary_models: bool = False) -> LineageSchema
...
         try:
             tests_results, invocation = self.tests_api.get_test_results(
                 filter=self.filter,
-                days_back=days_back,
-                disable_passed_test_metrics=disable_passed_test_metrics,
+                test_results_db_rows=test_results_db_rows,
                 disable_samples=self.disable_samples,
             )
-            tests_info = []
+            tests_metadata = []
```
elongl, line 254: I think it needs to be called `test_metadatas`.
The current name makes it sound like there's a single metadata for many tests.

elementary/monitor/api/tests/tests.py
```diff
@@ -87,222 +74,227 @@ def _get_invocation_from_filter(
...
+                    display_name=test_sub_type_display_name,
+                    metrics=sample_data,
+                    result_description=test_result_db_row.test_results_description,
+                )
+            elif test_result_db_row.test_type == "schema_change":
+                test_results = ElementaryTestResultSchema(
+                    display_name=test_sub_type_display_name.lower(),
+                    result_description=test_result_db_row.test_results_description,
+                )
+        return test_results
```
elongl, line 235: image

elementary/monitor/api/tests/schema.py
elongl: Does the `invocations_order` represent the `row_number` in the ordered result-set (`order by generated_at`)?
If so, why does the object itself need to be aware of its position in a list?
I think it makes more sense to pass around the ordered list when the order matters rather than store it on the object.

elementary/monitor/api/tests/tests.py
```diff
@@ -87,222 +74,227 @@ def _get_invocation_from_filter(
...
 
+    @staticmethod
+    def _get_failed_rows_count(test_result_db_row: TestResultDBRowSchema) -> int:
+        failed_rows_count = -1
+        if (
+            test_result_db_row.status != "pass"
+            and test_result_db_row.test_results_description
+        ):
+            found_rows_number = re.search(
+                r"\d+", test_result_db_row.test_results_description
```
elongl, line 198: Why don't we use the `failures` column?

elementary/monitor/api/tests/tests.py
```diff
@@ -10,66 +10,53 @@
...
+                disable_passed_test_metrics=disable_passed_test_metrics,
+            ),
         )
-        tests_metadata = (
+        test_results = (
             json.loads(run_operation_response[0]) if run_operation_response else []
         )
-        tests_metadata = [
-            TestMetadataSchema(**test_metadata) for test_metadata in tests_metadata
+        test_results = [
```
elongl, line 56: This should already be ordered and used rather than the `invocation_order`.

elementary/monitor/dbt_project/macros/get_test_results.sql
```diff
@@ -1,24 +1,25 @@
...
-        tests_in_last_chosen_days as (
-            select *,
-                  {{ elementary.datediff(elementary.cast_as_timestamp('detected_at'), elementary.current_timestamp(), 'day') }} as days_diff,
-                  row_number() over (partition by model_unique_id, test_unique_id, column_name, test_sub_type order by detected_at desc) as row_number
-                from test_results
-        ),
-
-        latest_tests_in_the_last_chosen_days as (
-            select * from tests_in_last_chosen_days where row_number = 1
+        test_results_with_ivocations_order as (
```
elongl, line 7: Typo.
Should maybe be called `ordered_test_results`.

elementary/monitor/api/tests/tests.py
```diff
@@ -87,222 +74,227 @@ def _get_invocation_from_filter(
...
-            )
-            tests_metrics = (
-                json.loads(run_operation_response[0]) if run_operation_response else {}
-            )
-
-        self.set_run_cache(key=TESTS_SAMPLE_DATA, value=tests_metrics)
-        return tests_metrics
-
-    def get_test_info_from_test_metadata(
+    def get_test_metadata_from_test_result_db_row(
```
elongl, line 79: image
image

elementary/monitor/api/tests/schema.py
IDoneShaveIt: Because we use `rank` method and not `row_number` method at the sql to order the invocations (we can have 2 different test results with 1 as the invocation order) we need to be aware of it 🙂

elementary/monitor/api/tests/tests.py
IDoneShaveIt: This is an old code when snowflake didn't return the real failures (affected rows).
So we can change it and use it, but that means that users that are using old snowflake adapter won't have a real number (only 1)

elementary/monitor/api/tests/tests.py
IDoneShaveIt: Explained on the schema why we can't relay on the order 🙂  (rank instead of row_number)

elementary/monitor/data_monitoring/data_monitoring_report.py
IDoneShaveIt: Yeah I tend to agree that we need to simplify the flow 🙂 
I think that the original idea was to have an "API" call for each screen, but now we can use a single call (specially when both API's uses the same query).
I want to do it as a part of a different PR because I want to add improvements to the json as well (stop duplicating tests metadata on the json for example) which will require changes at the UI as well.

elementary/monitor/data_monitoring/data_monitoring_report.py
IDoneShaveIt: Done :)

elementary/monitor/api/tests/schema.py
elongl: Thanks for explaining.
Can we rename this to `invocations_rank_index`?
I think we should clarify that this represents an index within a set.

IDoneShaveIt: Yeah sure!

PR #553, opened 2023-01-11
Ele 78 re add source freshness support

elongl:


elementary/cli/cli.py
```diff
@@ -14,32 +16,23 @@
...
-root_folder = os.path.join(os.path.dirname(__file__), "..")
-modules = ["monitor"]
-
 logger = get_logger(__name__)
 
 
 class ElementaryCLI(click.MultiCommand):
+    _CMD_MAP = {
+        "monitor": monitor,
+        "report": report,
```
elongl, line 25: We will now have `edr report` in addition to `edr monitor report`.

docs/quickstart/send-slack-alerts.mdx
```diff
@@ -119,8 +119,9 @@ vars:
 ## Alert on source freshness failures
 
 By default, alerts are sent on dbt tests, models and snapshots.
-You can also monitor source freshness by running `dbt run-operation upload_source_freshness` right after each execution of `dbt source freshness`.
+You can also monitor source freshness by running `edr run-operation upload-source-freshness` right after each execution of `dbt source freshness`.
```
haritamar, line 122: I think we should specify here that the two commands need to run from the same computer and from the same folder 
(that was also true before but maybe less obvious when we ask to run edr)

elongl: Definitely agree!

PR #583, opened 2023-01-18
add permissions to github_token

RoiTabach:


elongl:
I believe the [`permissions`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) key needs to be defined on either the workflow globally, or a specific job, not a step.
image
Thanks for noticing!

.github/workflows/remind-docs-and-tests.yml
```diff
@@ -9,8 +9,9 @@ jobs:
       - uses: wow-actions/auto-comment@v1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            permissions: read-all|write-all
```
RoiTabach, line 12: @elongl you mean something like 

```suggestion
      permissions: read-all|write-all
      - uses: wow-actions/auto-comment@v1
        with:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/remind-docs-and-tests.yml
```diff
@@ -6,11 +6,12 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
+      permissions: read-all|write-all
```
elongl, line 9: This is still under `steps`.
It needs to be placed right after the job's name.

elongl:
image
It also says so on the action running for this PR so you can tell if it works or not.

PR #585, opened 2023-01-18
ELE-187 fix test runs order

IDoneShaveIt:
- Fixed the order of the test runs
- Fixed test runs duplication
- Added test for duplicate test runs
- Added test for test runs order

### DIFF
--- a/elementary/monitor/api/tests/tests.py
+++ b/elementary/monitor/api/tests/tests.py
@@ -275,7 +275,7 @@ class TestsAPI(APIClient):
                     test_result_db_row.invocation_id
                     or test_result_db_row.test_execution_id
                 )
-                # Currently the way we flat test results causing that there is suplication in test incovation for each test.
+                # Currently the way we flat test results causing that there is duplication in test invocation for each test.
                 # This if statement checks if the invocation is already counted or not.
                 if invocation_id not in grouped_invocation_ids[elementary_unique_id]:
                     grouped_invocation_ids[elementary_unique_id].append(invocation_id)
--- a/elementary/monitor/api/tests/tests.py
+++ b/elementary/monitor/api/tests/tests.py
@@ -267,20 +267,28 @@ class TestsAPI(APIClient):
         self, test_result_db_rows: List[TestResultDBRowSchema]
     ) -> Dict[TestUniqueIdType, InvocationsSchema]:
         grouped_invocations = defaultdict(list)
+        grouped_invocation_ids = defaultdict(list)
         for test_result_db_row in test_result_db_rows:
             try:
                 elementary_unique_id = test_result_db_row.elementary_unique_id
-                grouped_invocations[elementary_unique_id].append(
-                    InvocationSchema(
-                        id=test_result_db_row.invocation_id
-                        or test_result_db_row.test_execution_id,
-                        time_utc=test_result_db_row.detected_at,
-                        status=test_result_db_row.status,
-                        affected_rows=self._parse_affected_row(
-                            results_description=test_result_db_row.test_results_description
-                        ),
-                    )
+                invocation_id = (
+                    test_result_db_row.invocation_id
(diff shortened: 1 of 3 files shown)


elementary/monitor/api/tests/tests.py
```diff
@@ -267,20 +267,28 @@ def _get_invocations(
...
-                        id=test_result_db_row.invocation_id
-                        or test_result_db_row.test_execution_id,
-                        time_utc=test_result_db_row.detected_at,
-                        status=test_result_db_row.status,
-                        affected_rows=self._parse_affected_row(
-                            results_description=test_result_db_row.test_results_description
-                        ),
-                    )
+                invocation_id = (
+                    test_result_db_row.invocation_id
```
elongl, line 275: Why is there an `or` expression here?
I think there will always be an `invocation_id` and a `test_execution_id` in `elementary_test_results`.

elementary/monitor/api/tests/tests.py
```diff
@@ -267,20 +267,28 @@ def _get_invocations(
...
+                    or test_result_db_row.test_execution_id
                 )
+                # Currently the way we flat test results causing that there is suplication in test incovation for each test.
```
elongl, line 278: duplication* incovation*

elementary/monitor/api/tests/tests.py
```diff
@@ -267,20 +267,28 @@ def _get_invocations(
...
+                # This if statement checks if the invocation is already counted or not.
+                if invocation_id not in grouped_invocation_ids[elementary_unique_id]:
+                    grouped_invocation_ids[elementary_unique_id].append(invocation_id)
```
elongl, line 281: I don't understand why is there both `grouped_invocations` and `grouped_invocation_ids`.

elongl (commented):
I don't exactly understand the problem nor the solution.
Could you perhaps share some context?
Not as familiar with this part of the code.

elementary/monitor/api/tests/tests.py
IDoneShaveIt: It's a backwards compatibility 🙂 
On older versions of our dbt package we didn't have invocation id at the `elementary_test_results` table, and because that is an incremental table, we have cases where invocation id is `None`.
I do prefer moving into invocation id and not test execution id, so I left it as a fall back (This is actually a part of an older PR).

elementary/monitor/api/tests/tests.py
IDoneShaveIt: Invocation ids is the list of the invocation ids we found for the test.
I use it as an helper list to understand if I already added the same invocation to the grouped invocations object for the relevant test 🙂 
(I find it easier to look for an id inside a list of ids, instead of looking for id inside a list of objects that has the key id).

If you think that this is not simpler, I don't mind changing it 🙂

elementary/monitor/api/tests/tests.py
elongl: Why not use `test_execution_id` for all?

elementary/monitor/api/tests/tests.py
elongl: I see, makes sense.

IDoneShaveIt:
@elongl sure thing!
There are 2 problems that we have with the test runs invocations currently:

1. The order of the invocations on the graph is based on the order of the elements on the report json (should be fixed on the UI as well, back for returning the data in the correct order will do).
2. There are some cases (like schema changes), that due to the flattening we do for the test results at elementary_test_results table, we show duplicate invocations at the graph (this is happening because we split the same test into multiple rows which has the same test id, therefore we count the invocation twice - working on a better solution at the package as well). - In this PR when creating the data for the test runs graphs, I check for each graph if it already contains the invocation.

Hope that I made it clearer 🙂

elongl (approved):
Thanks for explaining.

PR #596, opened 2023-01-19
Added a contribution guide.

elongl:


elongl:
@IDoneShaveIt 
So I thought about this too, but I didn't want to create additional friction at this point.
I think at some point this might be required to state ahead of time, but for the moment I think that the act we run the tests in the CI is sufficient.

PR #597, opened 2023-01-19
Update share-report-ui.mdx

Deathfireofdoom:
Just to clarify that it also generates the report.

Not sure if I need to make a PR, if not, just close this one.


Maayan-s:
Yes, PR is the way to go 👍 
We manage our docs as code through the awesome [Mintlify](https://mintlify.com/) platform so we could collaborate on it just like this. Also, this makes us update the docs as part of our own PRs.

Thank you for your contribution!

PR #604, opened 2023-01-21
verify that the dbt package is up-to-date in the internal project

haritamar:


elementary/monitor/dbt_project_utils.py
```diff
@@ -1,15 +1,62 @@
...
+        return package_path
+
+    legacy_package_path = os.path.join(_PACKAGES_PATH, _DBT_PACKAGE_NAME)
+    if os.path.exists(legacy_package_path):
+        return legacy_package_path
+
+    return None
+
+
+def dbt_package_up_to_date() -> bool:
```
elongl, line 31: 1. Just for extra clarity, append with `is_`?
2. Do you think we should make that validation generic for the `packages.yml` so that it'll also include the `dbt-utils` or other packages we might add in the future? Or is it an overhead for the time being?

elongl (commented):
Looks great, one comment.

elementary/monitor/dbt_project_utils.py
haritamar: I'll change(1)
re. (2( - I actually thought the same and started with implementing it like this, but then saw that annoyingly the version inside `dbt-project.yml` for `dbt-utils` was not consistent with the version that was installed :( :(

See this for example:
https://github.com/dbt-labs/dbt-utils/blob/0.9.6/dbt_project.yml

So changed it for now to be only on our package. I could also blacklist `dbt-utils` but it felt a bit pointless as it's the only package aside from our own :)

elongl: Ohh right.
Totally reasonable then.
We’ll have to be very on point about bumping the version ourselves.
Let’s merge!

PR #606, opened 2023-01-22
Fix event_freshness_anomalies examples

edbizarro:
## Fix event_freshness_anomalies examples

The `event_freshness_anomalies` doc section have the same example as the `freshness_anomalies` causing an error if the user copy/paste into their code:

```bash
dbt.exceptions.CompilationException: Compilation Error in test elementary_freshness_anomalies_<redacted>_updated_at__update_ts_dms_at (models/staging/<redacted>/schema.yml)                                                                                                               
  macro 'dbt_macro__test_freshness_anomalies' takes no keyword argument 'event_timestamp_column'
```

### DIFF
--- a/docs/guides/add-elementary-tests.mdx
+++ b/docs/guides/add-elementary-tests.mdx
@@ -159,7 +159,7 @@ version: 2
 models:
   - name: < model name >
     tests:
-      - elementary.freshness_anomalies:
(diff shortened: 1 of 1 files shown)


elongl:
@edbizarro 
Thanks a lot for this contribution!

PR #619, opened 2023-01-25
Ele 206 small hourly time buckets on the anomaly

elongl:


elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
```diff
@@ -1,18 +1,13 @@
-{%- macro get_test_rows_sample(test, test_results_query, test_type, results_sample_limit = 5) -%}
-    {% set result_rows = test.result_rows %}
-    {% if result_rows is defined and result_rows is not none %}
-      {% do return(fromjson(result_rows)) %}
-    {% endif %}
-
-    {% set test_rows_sample = none %}
-    {% if test_results_query %}
-        {% set test_rows_sample_query = test_results_query %}
+{%- macro get_test_rows_sample(result_rows_agate, test_type, results_sample_limit=5) -%}
```
elongl, line 1: Here there's a small compatibility issue.
Previously, `result_rows` had `null` if the results were too big to fit into the column.
Now that column is deprecated, so for past test results we're unable to know if we should run the test's query or not.

haritamar, line 1: I assume we're going for a minor bump, right? (as this code is not backwards compatible)

elementary/monitor/dbt_project/macros/alerts/get_test_alerts.sql
```diff
@@ -68,48 +67,44 @@
         where suppression_status = 'pending'
     {% endset %}
 
-    {% set alerts_agate = run_query(select_pending_alerts_query) %}
+    {% set alerts_agate = elementary.run_query(select_pending_alerts_query) %}
+    {% set test_result_rows_agate = elementary_internal.get_result_rows_agate(days_back) %}
```
haritamar, line 71: Tiny concern - there's a chance of a race here between the two queries.
I don't think it's a big deal (worst case the samples will be missing in this case), but just worth double checking it won't cause us to crash.

elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
elongl: Are you referring to our usage of `test_result_rows` model?

elementary/monitor/dbt_project/macros/alerts/get_test_alerts.sql
elongl: I think that given that the `select_pending_alerts_query` is built from the `alerts` model which is incremented within `edr`, and that `test_result_rows` is built within the dbt package, it should necessarily contain the result rows of what's returned from `alerts`.

What do you think?

haritamar: Makes sense to me

elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
haritamar: Yeah
Let's discuss quickly on slack

PR #626, opened 2023-01-29
Ele 82 slack test results summary

IDoneShaveIt:
- Created Slack summary message
- Improved data monitoring report send-report method 
- Created single data monitoring filter class


elementary/clients/s3/client.py
```diff
@@ -53,10 +54,66 @@ def send_report(
...
+                return f"http://{bucket_name}.{aws_s3_website_url}"
+
+            except Exception as ex:
+                logger.warning(f"Unable to get bucket website URL: {ex}.")
+                return None
+        else:
+            return None
+
+    @staticmethod
+    def _get_aws_s3_website_url_from_location(location: str) -> str:
```
IDoneShaveIt, line 85: This list was taken from AWS official docs.

elementary/clients/slack/slack_message_builder.py
```diff
@@ -136,3 +151,16 @@ def get_slack_status_icon(status: str) -> str:
 
     def get_slack_message(self) -> SlackMessageSchema:
         return SlackMessageSchema(**self.slack_message)
+
+    @staticmethod
+    def prettify_and_dedup_list(
```
IDoneShaveIt, line 156: Moved from alerts slack message builder to the common slack message builder

elementary/monitor/data_monitoring/data_monitoring_filter.py
```diff
@@ -0,0 +1,92 @@
...
+from elementary.clients.dbt.dbt_runner import DbtRunner
+from elementary.monitor.api.selector.selector import SelectorAPI
+from elementary.monitor.data_monitoring.schema import DataMonitoringFilterSchema
+from elementary.tracking.anonymous_tracking import AnonymousTracking
+from elementary.utils.log import get_logger
+
+logger = get_logger(__name__)
+
+
+class DataMonitoringFilter:
```
IDoneShaveIt, line 13: Moved from DataMonitoringAlerts and DataMonitoringReport

elementary/clients/slack/schema.py
```diff
@@ -10,3 +14,12 @@ class SlackMessageSchema(BaseModel):
...
     attachments: Optional[list] = None
     blocks: Optional[list] = None
+
+    @validator("attachments", pre=True)
+    def validate_attachments(cls, attachments):
+        if isinstance(attachments, list) and len(attachments) > 50:
+            logger.error(
+                f"Slack message attachments limit is 50. You have {len(attachments)} attachments. Attachments were removed fro the message"
+            )
+            return None
```
RoiTabach, line 24: If we have 50+ attachments, instead of sending them, we send _nothing_ ? Why not at least send the first 50?

IDoneShaveIt: I think that there is no reason that we will get to the point where the schema is the one that is removing the attachments.
This is more for handling bad usage while developing.
Login an error and removing the attachments is more noticeable and will make us attend the issue + I think there is no use of sending partial data that you can't control where the cut is (it can send weird messages).

As you can see, when sending the summary message, I handle this case and send different message when we pass the limit.

If you think that this is still an odd behaviour, please let me know and we can discuss it more 🙂

RoiTabach: I see- it's more defensive programing than a user facing behaviour
Makes sense

elementary/clients/slack/schema.py
```diff
@@ -10,3 +14,12 @@ class SlackMessageSchema(BaseModel):
     text: Optional[str] = None
     attachments: Optional[list] = None
     blocks: Optional[list] = None
+
+    @validator("attachments", pre=True)
+    def validate_attachments(cls, attachments):
+        if isinstance(attachments, list) and len(attachments) > 50:
+            logger.error(
+                f"Slack message attachments limit is 50. You have {len(attachments)} attachments. Attachments were removed fro the message"
```
haritamar, line 22:
```suggestion
                f"Slack message attachments limit is 50. You have {len(attachments)} attachments. Attachments were removed from the message"
```

elementary/monitor/cli.py
```diff
@@ -454,6 +454,18 @@ def report(
...
+    type=str,
+    default=None,
+    help='Disable functualities from the "send-report" command.\nCurrently only --disable html_attachment is supported.',
+)
+@click.option(
+    "--include",
+    type=str,
+    default=None,
+    help="Include additional information at the test results summary message.\nCurrently only --include descriptions is supported.",
+)
```
haritamar, line 468: Why don't we include descriptions by default? It's too detailed?

elementary/monitor/data_monitoring/report/data_monitoring_report.py
```diff
@@ -221,22 +250,69 @@ def send_report(
...
+
+    def send_test_results_summary(
+        self,
+        days_back: Optional[int] = None,
+        test_runs_amount: Optional[int] = None,
+        disable_passed_test_metrics: bool = False,
+        bucket_website_url: Optional[str] = None,
+        include_description: bool = False,
+    ) -> bool:
+        if self.slack_client:
```
haritamar, line 286: It's a bit confusing to check if `self.slack_client` is defined here, as the code does nothing if it's not.

I think it's clearer to do it in the calling code instead (and fail here if this function is called without a slack client being set).

elementary/monitor/fetchers/tests/tests.py
```diff
@@ -58,8 +59,78 @@ def get_all_test_results_db_rows(
         ]
         return test_results
 
+    def get_test_restuls_summary(
```
haritamar, line 62: fix typo

elementary/clients/s3/client.py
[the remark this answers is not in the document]
IDoneShaveIt: You are right about the presigned URL, but this is how we guide our users for now because anyone with the signed link can access the presigned URL, so we decided not to do it for now.
We can talk about this, but this is our behaviour since day 1 and I don't think this is part of the task scope 🙂

elementary/clients/gcs/client.py
```diff
@@ -43,22 +44,38 @@ def send_report(
...
             if self.tracking:
                 self.tracking.record_cli_internal_exception(ex)
-            return False
-        return True
+            return False, bucket_website_url
+        return True, bucket_website_url
+
+    def get_bucket_website_url(
+        self, bucket_name: str, destination_bucket: Optional[str] = None
+    ) -> Optional[str]:
```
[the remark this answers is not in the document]
IDoneShaveIt: You are right about the presigned URL, but this is how we guide our users for now because anyone with the signed link can access the presigned URL, so we decided not to do it for now.
We can talk about this, but this is our behaviour since day 1 and I don't think this is part of the task scope 🙂

elementary/monitor/cli.py
IDoneShaveIt: Yes 🙂 (product definition) + It is padding the attachments (which has small size limit)

PR #666, opened 2023-02-13
ELE-85: Group Alerts by Table

RoiTabach:


RoiTabach:
will re-open when the class structure is re-done

PR #670, opened 2023-02-13
Conditionally using 'result_rows' if it's defined.

elongl:
Previous PR for comparison - https://github.com/elementary-data/elementary/pull/619.

### DIFF
--- a/elementary/monitor/dbt_project/macros/get_test_results.sql
+++ b/elementary/monitor/dbt_project/macros/get_test_results.sql
@@ -38,7 +38,8 @@
             severity,
             status,
             days_diff,
-            invocations_rank_index
+            invocations_rank_index,
+            result_rows
         from ordered_test_results
         where invocations_rank_index <= {{ invocations_per_test }}
         order by elementary_unique_id, invocations_rank_index desc
@@ -57,7 +58,7 @@
             {% if not disable_passed_test_metrics %}
                 {% do elementary_tests_allowlist_status.append('pass') %}
             {% endif %}
-            {% set test_rows_sample = elementary_internal.get_test_rows_sample(test_result_rows_agate.get(test.id), test_type, metrics_sample_limit) %}
+            {% set test_rows_sample = elementary_internal.get_test_rows_sample(test.result_rows, test_result_rows_agate.get(test.id), test_type, test.test_results_query, metrics_sample_limit) %}
             {%- if (test_type == 'dbt_test' and status in ['fail', 'warn']) or (test_type != 'dbt_test' and status in elementary_tests_allowlist_status) -%}
                 {# Dimension anomalies return multiple dimensions for the test rows sample, and needs to be handle differently. #}
                 {# Currently we show only the anomalous for all of the dimensions. #}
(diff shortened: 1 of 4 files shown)


elementary/monitor/dbt_project/macros/get_test_results.sql
```diff
@@ -57,8 +58,8 @@
             {% if not disable_passed_test_metrics %}
                 {% do elementary_tests_allowlist_status.append('pass') %}
             {% endif %}
-            {% set test_rows_sample = elementary_internal.get_test_rows_sample(test_result_rows_agate.get(test.id), test_type, metrics_sample_limit) %}
             {%- if (test_type == 'dbt_test' and status in ['fail', 'warn']) or (test_type != 'dbt_test' and status in elementary_tests_allowlist_status) -%}
+                {% set test_rows_sample = elementary_internal.get_test_rows_sample(test.result_rows, test_result_rows_agate.get(test.id), test_type, test.test_results_query, metrics_sample_limit) %}
```
haritamar, line 62: What is the reason for moving it to be inside the "if"?
(and in general I'm a bit confused abou

haritamar, line 62: What is the reason for moving it to be inside the "if"?
(and in general not sure I understand the purpose of this "if" - seems like it only does logic specific to dimension anomalies?)

elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
```diff
@@ -1,13 +1,31 @@
-{%- macro get_test_rows_sample(result_rows_agate, test_type, results_sample_limit=5) -%}
-    {% set test_rows_sample = [] %}
+{%- macro get_test_rows_sample(result_rows, result_rows_agate, test_type, test_query, results_sample_limit=5) -%}
```
haritamar, line 1: The parameter names here are a bit confusing - not sure someone stumbling on this code will immediately understand what's the difference between `result_rows` and `result_rows_agate`.

Maybe the first should be something like `result_rows_from_json` or `legacy_result_rows`?
Either way I'd add a comment clarifying it.

elementary/monitor/dbt_project/macros/get_test_results.sql
elongl: Error tests are not supposed to be sampled.
The reason it worked so far is because [error tests](https://github.com/elementary-data/dbt-data-reliability/commit/8f312b1c369dd3608a3af9f7273ab0f36438590b#diff-7e7a0c1e6c6911e382b5c70e89456fce2aa74e4da96f1cd0295733468df8d563R29) default to [empty list `result_rows`](https://github.com/elementary-data/dbt-data-reliability/commit/3a8aca03fc126a6b13a7fc0a5bbf98ea812de3a7).

However, now a user that uses `test_result_rows`, this is what will happen on `get_test_rows_sample`:
1. `elementary_test_results.result_rows` will be `null`, so it'll progress.
2. `elementary_test_results.test_result_rows` will not contain any samples because the test materialization failed.
3. It'll attempt to run the query and crash because it's broken.
Point being, error tests should not be sampled.

Issue #676, opened 2023-02-14
EDR Monitor with no --select doesn't send alerts to slack


RoiTabach:
**Describe the bug**
When running edr monitor with no "filters", no slack alerts are sent.

**To Reproduce**
0. Run dbt-data-reliability e2e tests (or generate a DBT project with errors and fails in any way you like)
1. Run "edr monitor -st <token> -ch <channel> -t <target> --project-dir <dir>"
2. see no slack alerts
3. choose a model you know is failing, e.g. "error_model" in the e2e tests
4. run `edr monitor -st <token> -ch <channel> -t snowflake --days-back 7 --select model:<model_name>`
**Expected behavior**
without filters, slack alerts are sent, unfiltered.

**Screenshots**
Filter view from pdb, it's not None but it's internals are all Nones, when running without --select 
<img width="1728" alt="image" src="https://user-images.githubusercontent.com/25003091/218860789-bfa64a63-9f02-4a03-a35c-18ac171e9429.png">
if/elif which causes the problem: 
<img width="662" alt="image" src="https://user-images.githubusercontent.com/25003091/218861253-7232db03-193f-46ba-bf1f-f30ef779b0ca.png">

**Environment (please complete the following information):**
 - edr Version: 0.6.10, can be found by running `pip show elementary-data`
 - dbt package Version: 0.6.13, can be found in `packages.yml` file

**Additional context**
I think it's related to a thread in support : https://elementary-community.slack.com/archives/C02CTC89LAX/p1676383019819779

IDoneShaveIt:
Resolved on elementary python package version v0.7.1

PR #678, opened 2023-02-14
alert if empty selector

RoiTabach:
this fixes https://github.com/elementary-data/elementary/issues/676


RoiTabach:
fixed by Idan in a different PR

PR #679, opened 2023-02-15
fix alerts filter

IDoneShaveIt:


elementary/monitor/fetchers/alerts/alert_filters.py
```diff
@@ -18,24 +18,30 @@ def filter_alerts(
...
         filtered_alerts = _filter_alerts_by_model(alerts, filter)
     elif filter.owner:
         filtered_alerts = _filter_alerts_by_owner(alerts, filter)
     elif filter.node_names is not None:
         filtered_alerts = _filter_alerts_by_node_names(alerts, filter)
+    # If the filter contains a filter that we don't support, we want to return an empty list of alerts
+    elif filter.selector:
+        logger.error(
+            f"An unsupported alerts selector has been provided - {filter.selector}!\nNo alert has been sent!"
+        )
```
RoiTabach, line 44: I think desired behaviour in case of an invalid selector is "don't filter and just send all alerts forward", not "filter everything out"

PR #693, opened 2023-02-20
ELE-386 Slack summary message touchups

IDoneShaveIt:


tests/unit/monitor/data_monitoring/report/test_slack_report_summary_message_builder.py
```diff
@@ -28,22 +27,20 @@ def test_add_details_to_slack_alert_attachments_limit(test_results_summary):
...
     )
     assert ":small_red_triangle: *Failed tests*" in attachments_as_string
     assert ":warning: *Warning*" in attachments_as_string
-    assert ":wrench: *Schema changes*" in attachments_as_string
     assert ":exclamation: *Error*" in attachments_as_string
 
     # Over attachments limitation
     message_builder = SlackReportSummaryMessageBuilder()
-    message_builder._add_details_to_slack_alert((test_results_summary * 5)[0:38])
+    message_builder._add_details_to_slack_alert((test_results_summary * 5)[0:40])
```
RoiTabach, line 43: Do these magic numbers mean anything? 38 is the expected number of alerts? Maybe it can be repalced by `len()` of something?

RoiTabach: (not that it's new behavior, but while we're at it...)

IDoneShaveIt: I can move this number to a const if you think that this is better :)
The idea of this line is to make sure I generate the exact limit of the attachments.
The problem is that this method is adding more attachments as part of the message formatting which makes it hard to use len / indicative consts to make the test more "cleaner".

I'll see what I can do about it

RoiTabach (approved):

PR #700, opened 2023-02-23
Ele 352 - Report performance

IDoneShaveIt:
- Created SlimDbtRunner that currently only support `run_operation` method
- Changed all the macros to return values instead of log them
- Added a mechanism `should_log` to `DbtRunner` that wraps the `run_operation` with the a macro that `edr_log` that logs the given macro results (It set to True by default and keeps the same behaviour as before the change)


elementary/clients/api/api_client.py
```diff
@@ -1,14 +1,15 @@
...
 from elementary.clients.dbt.dbt_runner import DbtRunner
+from elementary.clients.dbt.slim_dbt_runner import SlimDbtRunner
 from elementary.utils.log import get_logger
 
 logger = get_logger(__name__)
 
 
 class APIClient:
-    def __init__(self, dbt_runner: DbtRunner):
+    def __init__(self, dbt_runner: Union[DbtRunner, SlimDbtRunner]):
```
haritamar, line 12: Why not change the type to be simply `BaseDbtRunner`?

elementary/clients/dbt/slim_dbt_runner.py
```diff
@@ -0,0 +1,177 @@
+# flake8: noqa
+import json
+from typing import Optional
+
+import dbt.adapters.factory
+
+dbt.adapters.factory.get_adapter = lambda config: config.adapter
```
haritamar, line 7: Maybe add a comment here that this needs to be kept before other dbt imports

elementary/clients/fetcher/fetcher.py
```diff
@@ -1,11 +1,12 @@
...
-from typing import Any
+from typing import Any, Union
 
 from elementary.clients.dbt.dbt_runner import DbtRunner
+from elementary.clients.dbt.slim_dbt_runner import SlimDbtRunner
 
 
 class FetcherClient:
-    def __init__(self, dbt_runner: DbtRunner):
+    def __init__(self, dbt_runner: Union[DbtRunner, SlimDbtRunner]):
```
haritamar, line 9: also here re. the type

elementary/clients/api/api_client.py
IDoneShaveIt: IMO BaseDbtRunner is too generic.
Here I know exactly which DbtRunners are supported, but it is not that matters currently because we support all :)

haritamar: I think that this class can actually support any class that inherits from BaseDbtRunner?
E.g. if I want to initialize it with ItamarCustomDbtRunner that implements the same methods, it should work, no?

I think it makes sense to specify the minimal interface that is required, that is the point of inheritance :)

PR #716, opened 2023-02-27
ELE-393 support bucket-website-url cli param

IDoneShaveIt:


elementary/clients/gcs/client.py
```diff
@@ -71,12 +73,15 @@ def get_bucket_website_url(
...
-            bucket_website_url = f"https://storage.googleapis.com/{full_bucket_path}"
+            if self.config.bucket_website_url:
+                bucket_website_url = self.config.bucket_website_url
+            else:
+                full_bucket_path = (
+                    f"{destination_bucket}/{bucket_name}"
+                    if destination_bucket
+                    else bucket_name
+                )
+                bucket_website_url = f"{DEFAULT_BUCKET_WEBSITE_URL}/{full_bucket_path}"
```
elongl, line 84: I think we should use [`urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin).

elementary/clients/s3/client.py
```diff
@@ -64,22 +64,25 @@ def send_report(
...
+                    )
+                    bucket_website_url = f"http://{bucket_name}.{aws_s3_website_url}"
 
-            except Exception as ex:
-                logger.warning(f"Unable to get bucket website URL: {ex}.")
-                return None
-        else:
-            return None
+                except Exception as ex:
+                    logger.warning(f"Unable to get bucket website URL: {ex}.")
```
elongl, line 83: Why do we catch this?
If `send-report` failed to send a report it should crash.

IDoneShaveIt: So this is the current behaviour and I think that the idea behind it is that you can upload the report and send it to Slack at the same time.
So if one is failing, we don't want to fail the other.
We can change it but I am not sure if this is part of the task context 🤔

PR #738, opened 2023-03-06
ELE-485 remove user DWH query for test result samples

IDoneShaveIt:
Reverting the changes of https://github.com/elementary-data/elementary/pull/736, and removing the backcomp support for fetching test result samples from the user DWH

### DIFF
--- a/elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
+++ b/elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
@@ -26,5 +26,5 @@
         {{ return(result_rows) }}
     {% endif %}
 
-    {% do return(elementary.agate_to_dicts([])) %}
+    {% do return([]) %}
 {%- endmacro -%}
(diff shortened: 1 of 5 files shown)


elementary/monitor/dbt_project/macros/get_test_rows_sample.sql
```diff
@@ -26,15 +26,5 @@
...
-        with test_results as (
-          {{ test_query }}
-        )
-        select * from test_results {% if should_limit_sample %} limit {{ results_sample_limit }} {% endif %}
-      {% endset %}
-      {% do return(elementary.agate_to_dicts(elementary.run_query(query))) %}  
-    {% else %}
-      {% do return(elementary.agate_to_dicts([])) %}
-    {% endif %}
+    {% do return(elementary.agate_to_dicts([])) %}
```
elongl, line 29: Why do we use `agate_to_dicts` here instead of `return([])`?

IDoneShaveIt: No real reason, a changed I did when checked it and forgot it like that.
Changing

PR #743, opened 2023-03-09
make DataMonitoring OK with getting None tracking

RoiTabach:
This will help when calling those classes from Airflow - we don't pass tracking there and don't have access to MockTracking from elementary/tests


IDoneShaveIt (commented):
Please change the typing as well to Optional[AnonymousTracking] on all of the data monitoring files 🙂

PR #744, opened 2023-03-09
Allow `tracking` to be Optional

RoiTabach:


elongl:
I think a better approach would be to enable the [`_do_not_track` property](https://github.com/elementary-data/elementary/blob/29eef4c29c2aea34cad0a642bdfb551a27908c83/elementary/tracking/anonymous_tracking.py#L37-L38) on the tracking object because it won't require doing `if self.tracking` whenever we use the tracking.

elongl (approved):

PR #747, opened 2023-03-13
Add target_path flag option for edr files and logs

noaKurman:


RoiTabach:
Looks good, added one small comment 
Do we want to add user-facing documentation for it?

Maayan-s:
> Looks good, added one small comment
> Do we want to add user-facing documentation for it?

@noaKurman 
If you don't document no one will use it, so it's not a question 🙃

noaKurman:
> > Looks good, added one small comment
> > Do we want to add user-facing documentation for it?
> 
> @noaKurman If you don't document no one will use it, so it's not a question 🙃 Let me know if you need help with docs. You can do it in different PR
@Maayan-s @RoiTabach 
I added documentation in docs/understand-elementary/cli-commands.mdx, is that enough? If not I'll add it in a different PR :)

docs/understand-elementary/cli-commands.mdx
```diff
@@ -61,6 +61,13 @@ Here is a list of all the available options:
 - **Configuration directory path:**
   The path of the directory in which your `config.yml` file is located. Default is `HOME_DIR/.edr`.
 
+```shell
+-tp, --target-path <PATH>
+```
+
+- **Configuration target directory path**
+  The path where all output files will be saved, default will create "edr_files" directory in your path
```
Maayan-s, line 69: @noaKurman 
Can they change the default? 
If not I would change to:
The path where `edr` should create an `edr_files` directory where all output files will be saved.

noaKurman: Not completely sure what you meant:
You can either run `--target-path ./noa-target` and it would create `noa-target` in the current directory and save all files there, or you can ignore the flag and it would create `edr_files` in your current directory and save all files there, so should I change it to the phrase you suggested? 
Also should I mention it's an absolute path in the doc?
@Maayan-s

Maayan-s: Ok I get it now! 

So I would write like this:
The absolute path where all output files will be saved. If not configured, the default is `./edr_files`.

elementary/config/config.py
```diff
@@ -69,10 +72,12 @@ def __init__(
 
         config = self._load_configuration()
 
-        self.target_dir = self._first_not_none(
+        self.target_dir = self.target_dir = self._first_not_none(
```
RoiTabach, line 75: Typo without side effects but still - let's fix: 

```suggestion
        self.target_dir = self._first_not_none(
```

elementary/utils/log.py
```diff
@@ -25,7 +28,30 @@ def format(self, record):
...
+            ctx = click.get_current_context()
+            ctx_args = ctx.args
+            if target_path_flag in ctx_args:
+                file_path = ctx_args[ctx_args.index(target_path_flag) + 1]
+            elif target_path_param in ctx.params:
+                file_path = ctx.params[target_path_param]
+        except Exception:
+            pass
+        finally:
+            os.makedirs(os.path.abspath(os.path.abspath(file_path)), exist_ok=True)
```
RoiTabach, line 49: we don't need to double `abspath` , right? 
```suggestion
            os.makedirs(os.path.abspath(file_path), exist_ok=True)
```

elementary/utils/log.py
```diff
@@ -25,7 +28,30 @@ def format(self, record):
 
 
 FORMATTER = ColoredFormatter()
-LOG_FILE = "edr.log"
+LOG_FILE = os.getcwd() + "/edr_files/edr.log"
+
+
+class CustomerFileHandler(logging.FileHandler):
```
RoiTabach, line 34: Is this `customer` and not `custom` intentionaly? I don't know English well enough to be sure

elementary/utils/log.py
```diff
@@ -36,7 +62,7 @@ def get_console_handler():
 
 
 def get_file_handler():
-    file_handler = logging.FileHandler(LOG_FILE)
+    file_handler = CustomFileHandler(LOG_FILE, delay=True)
```
haritamar, line 65: I think we should add the file handler dynamically with the correct log file output, initializing it with `LOG_FILE` and then changing it in `emit` feels a bit hacky...

I think it's better that we'll add a line of the format:
```logger.addHandler(get_file_handler(log_path))```

to `elementary/monitor/cli.py` (where `log_path` is constructed from the target path parameter.

File: elementary/utils/log.py [hunk 6]
```diff
@@ -25,7 +28,30 @@ def format(self, record):
...
+
+class CustomerFileHandler(logging.FileHandler):
+    def emit(self, record):
+        file_path = os.getcwd() + "/edr_files"
+        try:
+            target_path_param = "target_path"
+            target_path_flag = "--target-path"
+            ctx = click.get_current_context()
+            ctx_args = ctx.args
+            if target_path_flag in ctx_args:
```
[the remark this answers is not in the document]
noaKurman: Because in the context of the cli/cli.py it's an arg, but in the context of some other files it's in the params. I guess its based on if it's before or after the `get_cli_properties` function @haritamar

elementary/monitor/cli.py
```diff
@@ -154,10 +158,12 @@ def get_cli_properties() -> dict:
         return dict()
 
     reload_monitoring_configuration = params.get("reload_monitoring_configuration")
+    target_path = params.get("target_path")
     update_dbt_package = params.get("update_dbt_package")
     full_refresh_dbt_package = params.get("full_refresh_dbt_package")
 
     return {
+        "target_path": target_path,
```
elongl, line 166: Is this necessary?
What does it affect?

elementary/config/config.py
```diff
@@ -31,13 +31,16 @@ class Config:
 
     DEFAULT_CONFIG_DIR = str(Path.home() / ".edr")
 
+    DEFAULT_FILES_PATH = os.getcwd() + "/edr_files"
```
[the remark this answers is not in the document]
noaKurman: Or asked for "edr_target" since it's compliant with dbt target dir

elementary/monitor/cli.py
noaKurman: Are you asking if we need it in the return or if we need the `"target_path": target_path` instead of just returning target_path itself?

For the first question - I'm pretty sure that's how we set the params (instead of it being in the args)

elementary/utils/log.py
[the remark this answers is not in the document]
noaKurman: Should have been Custom 😅

elementary/utils/log.py
```diff
@@ -46,7 +61,10 @@ def get_logger(logger_name):
     logger = logging.getLogger(logger_name)
     logger.setLevel(logging.DEBUG)
     logger.addHandler(get_console_handler())
-    logger.addHandler(get_file_handler())
     # with this pattern, it's rarely necessary to propagate the error up to parent
-    logger.propagate = False
     return logger
+
+
+def set_root_logger_file_handler(logger_name):
```
haritamar, line 68: I think you should add both handlers only to the root logger, otherwise you'll get duplicate console outputs. 
So maybe we can have `set_root_logger_handlers`?

elementary/cli/cli.py
```diff
@@ -45,6 +45,7 @@ def format_help(self, ctx, formatter):
         self.format_epilog(ctx, formatter)
 
     def invoke(self, ctx: click.Context) -> Any:
+        set_root_logger_file_handler("elementary")
```
haritamar, line 48: I think it may be good that `set_root_logger_file_handler` will explicitly get the target path rather than relying on the click context.
WDYT about moving the call to `config.py` below and then doing `set_root_logger_file_handler("elementary", target_path)`?

I think it makes sense to do it right after you're creating the directory there.

elementary/utils/log.py
```diff
@@ -35,8 +37,21 @@ def get_console_handler():
     return console_handler
 
 
+def get_log_path():
```
haritamar, line 40: See my note above regarding getting the target_path as a parameter, I think it can simplify this function

elementary/utils/log.py
```diff
@@ -35,8 +37,21 @@ def get_console_handler():
...
+    file_path = os.getcwd() + "/edr_target/"
+    try:
+        ctx = click.get_current_context()
+        ctx_args = ctx.args
+        target_path_flag = "--target-path"
+        file_path = ctx_args[ctx_args.index(target_path_flag) + 1]
+    finally:
+        os.makedirs(os.path.abspath(file_path), exist_ok=True)
+        return file_path + "/edr.log"
```
haritamar, line 49: it's a good practice in python usually to use `os.path.join(file_path, "edr.log")` rather than adding strings 
(handles well edge cases, like if the path ends or doesn't end with "/")

.github/workflows/test-github-action.yml
```diff
@@ -100,4 +100,4 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: edr.log
-          path: edr.log
+          path: edr_files/edr.log
```
haritamar, line 103: Should this be changed to `edr_target`? (also for other cases below)

elementary/utils/log.py
noaKurman: Ha good to know, I saw both in the code so just used one of them 👍

elementary/utils/log.py
noaKurman: I actually don't see it log twice to the console so it's interesting
But anyway I'll move the console handler to the root as well, I agree it's more sorted that way

elementary/cli/cli.py
noaKurman: Hmm I see at that point in the code the config still points to the default edr_files path, maybe you can show me why 
but I can also get it from the click context in the cli.py instead of the log (unless what you dislike here is the context itself and not the location of it)

PR #753, opened 2023-03-15
Fixing incorrect group_by flag documentation for slack

vishaalkk:
This PR fixes incorrect documentation about the group_by flag in Slack alerts. CLI takes the flag as `--group-by` instead of `--group_by`.

image


Maayan-s:
Thank you for fixing @vishaalkk! 🤩

PR #766, opened 2023-03-21
Ele 527 high memory consumption when running edr

oravi:


elementary/clients/dbt/dbt_runner.py
```diff
@@ -71,7 +72,9 @@ def _run_command(
         output = None
         if capture_output:
             output = result.stdout.decode("utf-8")
-            logger.debug(f"Output: {output}")
+            logger.debug(
```
elongl, line 75: As having the output in the log was pretty useful so far, do you think it should consider [`is_debug()`](https://github.com/elementary-data/elementary/blob/29eef4c29c2aea34cad0a642bdfb551a27908c83/elementary/utils/env_vars.py#L4) so that we can tell users to run `EDR_DEBUG=1 edr ...` and then send us the `edr.log`? Or do `edr --debug ...`.

elongl (approved):

PR #771, opened 2023-03-22
fix issue in parsing of subscribers

RoiTabach:
we expected an empty list but subscribers can be sometimes `None`

### DIFF
--- a/elementary/utils/json_utils.py
+++ b/elementary/utils/json_utils.py
@@ -63,6 +63,7 @@ def unpack_and_flatten_and_dedup_list_of_strings(
 def list_of_lists_of_strings_to_comma_delimited_unique_strings(
     list_of_strings: List[List[str]], prefix: Optional[str] = None
 ) -> str:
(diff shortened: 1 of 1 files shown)


IDoneShaveIt (approved):

elementary/utils/json_utils.py
```diff
@@ -63,6 +63,7 @@ def unpack_and_flatten_and_dedup_list_of_strings(
 def list_of_lists_of_strings_to_comma_delimited_unique_strings(
     list_of_strings: List[List[str]], prefix: Optional[str] = None
 ) -> str:
+    list_of_strings = [x for x in list_of_strings if x] # filter Nones and empty lists
```
IDoneShaveIt, line 66: Why would we have an empty string / None in the list?

RoiTabach, line 66:
```suggestion
    list_of_strings = [x for x in list_of_strings if x]  # filter Nones and empty lists
```

PR #773, opened 2023-03-23
Fix config and target directory path docs

will-warner:
The flags for setting both the config and target paths are the wrong way around.

Correcting the order here.


elongl:
Thanks a lot!

PR #778, opened 2023-03-26
Add a precision to "What happens on each test"

winzee:
The precision on the behavior was confirmed by Or Avidov on Slack, and he asked me to correct it.


Maayan-s:
Thank you @winzee !

PR #783, opened 2023-03-28
Add log rotation + dbt internal project log target

noaKurman:
I tried first using the cmd flag, but it requires passing the target path from locations where it is set, and it's much less elegant. 
I can explain further if there's an issue with setting the env var to solve this

### DIFF
--- a/elementary/cli/cli.py
+++ b/elementary/cli/cli.py
@@ -27,7 +27,6 @@ def get_log_path(ctx):
         target_path = ctx_args[ctx_args.index(target_path_flag) + 1]
     finally:
         os.makedirs(os.path.abspath(target_path), exist_ok=True)
-        os.environ["DBT_LOG_PATH"] = os.path.abspath(target_path)
         return os.path.join(target_path, "edr.log")
 
 
--- a/elementary/cli/cli.py
+++ b/elementary/cli/cli.py
@@ -27,6 +27,7 @@ def get_log_path(ctx):
         target_path = ctx_args[ctx_args.index(target_path_flag) + 1]
     finally:
         os.makedirs(os.path.abspath(target_path), exist_ok=True)
+        os.environ["DBT_LOG_PATH"] = os.path.abspath(target_path)
         return os.path.join(target_path, "edr.log")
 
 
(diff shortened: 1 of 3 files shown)


elementary/utils/log.py
```diff
@@ -25,6 +26,8 @@ def format(self, record):
 
 
 FORMATTER = ColoredFormatter()
+MAX_BYTES_IN_FILE = 10240000
```
noaKurman, line 29: Pretty arbitrary choice from some examples I saw. No hard feelings if you want me to change that

elementary/utils/log.py
```diff
@@ -51,3 +61,4 @@ def set_root_logger_handlers(logger_name, files_target_path):
     logger = logging.getLogger(logger_name)
     logger.addHandler(get_file_handler(files_target_path))
     logger.addHandler(get_console_handler())
+    logger.addHandler(log_rotation_handler(files_target_path))
```
elongl, line 64: Shouldn't we replace the `logging.FileHandler` with `logging.RotatingFileHandler` in `get_file_handler` instead?
Now it seems that there are two file handlers and both are writing to the same file.

elementary/utils/log.py
elongl: Maybe we can [copy dbt](https://github.com/dbt-labs/dbt-core/blob/9bb1250869d6d004f2a81656986bd03c03cf12da/core/dbt/logger.py#L357).

elementary/cli/cli.py
```diff
@@ -27,6 +27,7 @@ def get_log_path(ctx):
         target_path = ctx_args[ctx_args.index(target_path_flag) + 1]
     finally:
         os.makedirs(os.path.abspath(target_path), exist_ok=True)
+        os.environ["DBT_LOG_PATH"] = os.path.abspath(target_path)
```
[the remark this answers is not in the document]
noaKurman: the target_path isn't available from anywhere in the code since it only exist when we have the context and some functions initialize before this happens (for example I can't place it in the init of the dbtRunner because it's being imported before, and placing it inside the functions themselves doesn't feel fit) 
But you're right it doesn't really belong in this function 
I'm think of placing it in the config, but I need to make sure the config will be initialized before the dbt is running anywhere 👍

elementary/utils/log.py
noaKurman: It's pretty much the same (10240000  is also around 10 mb). But theirs is more readable so I'll change it :)

ele-elon (approved):

PR #791, opened 2023-03-30
indicative incompatibility error

Maayan-s:


elementary/monitor/data_monitoring/data_monitoring.py
```diff
@@ -123,17 +123,31 @@ def get_latest_invocation(self) -> Dict[str, Any]:
...
         ):
             logger.warning(
                 f"You are using incompatible versions between edr ({py_pkg_ver}) and Elementary's dbt package ({dbt_pkg_ver}).\n "
-                "Please upgrade the major and minor versions to align.\n",
+                "To fix please update your packages.yml, and run:\n"
+                "dbt deps && dbt run --select elementary\n",
             )
+            return
+
+        logger.info(
```
elongl, line 151: Maybe we should just drop this log?
I tend to think that it's more confusing to the user rather than valuable information.

elementary/monitor/data_monitoring/data_monitoring.py
```diff
@@ -123,17 +123,31 @@ def get_latest_invocation(self) -> Dict[str, Any]:
...
-            dbt_pkg_ver.major != py_pkg_ver.major
-            or dbt_pkg_ver.minor != py_pkg_ver.minor
+        if dbt_pkg_ver.major > py_pkg_ver.major or (
+            dbt_pkg_ver.major == py_pkg_ver.major
+            and dbt_pkg_ver.minor > py_pkg_ver.minor
+        ):
+            logger.warning(
+                f"You are using incompatible versions between edr ({py_pkg_ver}) and Elementary's dbt package ({dbt_pkg_ver}).\n "
+                "To fix please run:\n"
+                "pip install elementary-data --upgrade\n",
```
elongl, line 136: > pip install [options] <requirement specifier> [package-index-options] ...

The `--upgrade` needs to be before the package name.
```suggestion
                "pip install --upgrade elementary-data\n",
```

elementary/monitor/data_monitoring/data_monitoring.py
Maayan-s: I thought about this, but it could help us when debugging problems for users. 
Also, I think that when they run the tool it's overall positive that they see progress and don't think it's stuck.

PR #792, opened 2023-03-30
Fix typo in docs

jkwalsh05:
back_days referenced when it should be backfill_days

### DIFF
--- a/docs/guides/add-elementary-tests.mdx
+++ b/docs/guides/add-elementary-tests.mdx
@@ -85,7 +85,7 @@ models:
 Monitors the freshness of your table over time, as the expected time between data updates.
 
 Upon running the test, your data is split into time buckets (daily by default, configurable with the `time bucket`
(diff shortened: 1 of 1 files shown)


elongl:
Hi @jkwalsh05, thanks for suggesting this change.
Actually, those are two different configuration values.
Unfortunately the naming is not clear enough and doesn't indicate its true meaning,
but you can roughly view `days_back` as the training period for which we compare data to, and `backfill_days` as the test period that is compared against the training period. Additionally, you'd receive failure in tests of anomalies within that period. I'm going to close this PR for now, but I definitely think we should invest into explaining the test mechanism a bit better. However, we are planning on simplifying it in the near future so it might be a good time to do it then.

elongl:
I do see though that the [*other* reference is invalid](https://github.com/elementary-data/elementary/blob/7f8da18d778967c9a05325c0e5d1591dcd7a1c64/docs/guides/add-elementary-tests.mdx#L28) and should be `backfill_days`.
Would you please like to contribute that?

PR #837, opened 2023-04-27
Ele 381 add mypy to the cli repo - step 1

haritamar:


mypy.ini
```diff
@@ -0,0 +1,16 @@
+[mypy]
+packages = elementary
+exclude = ^elementary/monitor/dbt_project/
+allow_redefinition = True
```
haritamar, line 4: This means you can do something like:
```
res = [... some list comprehension ... ].    # Type is List[str]
res = json.dumps(res)                                # Type is str - mypy by default alerts when that happens
```

I thought in some cases it is legitimate and online that other people also allowed this in their projects. But we can also be more strict

mypy.ini
```diff
@@ -0,0 +1,16 @@
...
+
+[mypy-alive_progress]
+ignore_missing_imports = True
```
haritamar, line 7: For every library that didn't have a "stubs" library I needed to do this (See the libs I added to `dev-requirements.txt`)

elementary/monitor/api/alerts/alerts.py
```diff
@@ -146,26 +137,25 @@ def _sort_alerts(
...
+        alerts: AlertsQueryResult[ValidAlertType],
         last_alert_sent_times: Dict[str, str],
     ) -> List[str]:
         suppressed_alerts = []
         current_time_utc = datetime.utcnow()
-        for alert in [*alerts.alerts, *alerts.malformed_alerts]:
-            alert_class_id = alert.alert_class_id
+        all_alerts: List[Alert] = [*alerts.alerts, *alerts.malformed_alerts]
+        for alert in all_alerts:
+            alert_class_id = alert.alert_class_id or ""
```
haritamar, line 155: So I did this in a bunch of places
Because in many classes we have fields defined as optional, where practically they are not really optional (or not optional in specific flows)
I wasn't sure if it's safe to change them not to be optional (can cause pydantic to fail in runtime when we initialize classes from dbt results)

elementary/monitor/api/filters/filters.py
```diff
@@ -50,7 +51,8 @@ def _get_test_filters(
 
         totals_models_ids = totals.keys()
         for artifact in [*models.values(), *sources.values()]:
-            if artifact.unique_id not in totals_models_ids:
+            assert isinstance(artifact, ArtifactSchema)
```
haritamar, line 54: I added asserts in many places where I needed to [narrow down the type](https://mypy.readthedocs.io/en/stable/type_narrowing.html) for mypy

I think in this case it's fine to use asserts as I'm not relying on them as production checks

elongl: Wouldn't it be preferable to use [`cast()`](https://mypy.readthedocs.io/en/stable/type_narrowing.html#casts) in that case?

mypy.ini
elongl: I think that we shouldn't allow this.
Just confusing in my opinion and better to call the new variable `res_dump` for instance.
It highlights the change and is easier to understand.

elementary/clients/slack/slack_message_builder.py
```diff
@@ -155,14 +155,12 @@ def get_slack_message(self) -> SlackMessageSchema:
...
-            return ", ".join(set(list_variation))
-
-        else:
-            return list_variation
+    def prettify_and_dedup_list(str_list: Union[List[str], str]) -> str:
+        """
+        Receives a list of strings, either JSON dumped or not, dedups and sorts it, and returns it as a comma-separated
+        string.
+        This is useful for various lists we include in slack messages (owners, subscribers, etc)
+        """
```
haritamar, line 163: I removed `prettify_json_str_set` because it was confusing.
Made this function always return a string because that's the intention (it's for the owner list in slack messages)

mypy.ini
haritamar: Hmm ok I'll try to remove it and see how many cases it produces

elementary/monitor/alerts/alerts.py
```diff
@@ -22,9 +22,9 @@ def count(self) -> int:
         return len(self.alerts) + len(self.malformed_alerts)
 
     def get_all(self) -> List[Alert]:
```
elongl, line 24: Isn't there a way to say that it's okay given that they both inherit from `Alert`?

elementary/monitor/alerts/alert.py
```diff
@@ -80,6 +85,15 @@ def consice_name(self):
         return "Alert"
 
 
+class ValidAlert(Alert):
```
elongl, line 88: What's the purpose of this?

elementary/monitor/alerts/group_of_alerts.py
```diff
@@ -214,6 +214,7 @@ def _get_model_error_block_header(self) -> List:
             return []
         result = []
         for model_error_alert in model_error_alert_list:
+            assert isinstance(model_error_alert, ModelAlert)
```
elongl, line 217: How could a non `ModelAlert` be in this list in the first place?
Maybe we should make a check and raise an error?

elementary/monitor/alerts/alert.py
elongl, line 88: Care to clarify what's the purpose of this compared to `Alert`?

elementary/monitor/alerts/alerts.py
haritamar: solved - I believe I did this before I added a bound to AlertType

elementary/monitor/alerts/group_of_alerts.py
haritamar: The problem is that `self._components_to_alerts` is a dict whose value changes according to the provided key. There's no way that I can tell to make mypy understand it.
Changed to an exception though.

haritamar: (may worth refactoring, but I want to avoid it for now)

elementary/monitor/alerts/alert.py
haritamar: Removed it. 
I added it only to simplify some of the Unions we did, but didn't actually need it.

elementary/monitor/api/alerts/alerts.py
```diff
@@ -66,7 +67,7 @@ def get_test_alerts(
         test_alerts = self._sort_alerts(
             pending_test_alerts, last_alert_sent_times, filter
         )
-        return test_alerts
+        return cast(AlertsQueryResult[TestAlert], test_alerts)
```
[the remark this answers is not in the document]
haritamar: generics were the solution.
I didn't realize this allows matching types between the parameters and and the result, which is exactly what was needed here.

elongl (approved):

PR #851, opened 2023-05-08
Added support for --dbt-vars in report command

### LINKED ISSUE #793: Add support for --dbt-vars at the report CLI commands
Some of the users' profiles depends on dbt vars.
Currently the CLI supports passing dbt vars only for the monitor commands, making the report commands not work.

Add a support for --dbt-vars to the rest of the CLI commands.
The dbt runner is already supporting it, we should just pass it to the report CLI commands as well.

manulpatel:
Closes #793 

Added support for --dbt-vars parameter in report command in elementary cli.


elementary/monitor/cli.py
```diff
@@ -352,6 +359,7 @@ def report(
     """
     Generate a local observability report of your warehouse.
     """
+    vars = yaml.loads(dbt_vars) if dbt_vars else None
```
elongl, line 362: Can you rename it to `dbt_vars` or directly assign it as the argument?
This overrides the `vars` function.

elementary/monitor/cli.py
```diff
@@ -383,6 +391,7 @@ def report(
             exclude_elementary_models=exclude_elementary_models,
             should_open_browser=open_browser,
             project_name=project_name,
+            dbt_vars=vars,
```
elongl, line 394: This function does not accept `dbt_vars`.
I think we should add an `dbt_env_vars` parameter to [`Config`](https://github.com/elementary-data/elementary/blob/fe8751be51021f3c2eeca165d25768b21a17dd80/elementary/config/config.py#L70) and pass it there. At the moment, it only creates one internally.

If we make that change, I think we should also remove `dbt_vars` from [`run_alerts`](https://github.com/elementary-data/elementary/blob/90142313e187d3d2f236015826cfdf34c692b466/elementary/monitor/cli.py#L291) and instead read it from the `Config` above.

elongl (commented):
Thanks a lot for submitting the PR.
Left a few comments.

elementary/monitor/cli.py
manulpatel: Okay got it. Inside the `__init__` of config.py what should be the type of ` dbt_vars` string or dict? Currently it is empty dict but [here ](https://docs.getdbt.com/reference/dbt-jinja-functions/var) it seem string.

Also in [run_alerts](https://github.com/elementary-data/elementary/blob/90142313e187d3d2f236015826cfdf34c692b466/elementary/monitor/cli.py#L291), do we really need to remove `dbt_vars` from it? Because I think in this issue we are only defining new option for `report` command and not for `monitor`.

elementary/monitor/cli.py
elongl: Yes, sounds great.

elementary/monitor/cli.py
manulpatel: Yes seems fine.

NoyaArie (commented):
Thank you for contributing this feature 💪 
I recommend adding a new small test that covers the case that dbt_vars have a value. This will make sure this feature won't break by future changes :)

elementary/config/config.py
```diff
@@ -67,7 +68,7 @@ def __init__(
         self.env = env
 
         # Additional env vars supplied to dbt invocations
-        self.dbt_env_vars = dict()
+        self.dbt_env_vars = dbt_vars
```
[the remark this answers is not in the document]
manulpatel: Thanks for poinits out. Updated in new commit.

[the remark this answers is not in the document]
manulpatel: @NoyaArie Now all exisiting tests are passing. Do I need to write new unit tests?

elementary/monitor/cli.py
IDoneShaveIt: Hey @manulpatel, chiming in here 🙂
I just want to clarify and make sure you aware that this change will load the vars as `env vars`.
The meaning of this is that in order to access them you will have to use [env_var](https://docs.getdbt.com/reference/dbt-jinja-functions/env_var) method instead of the [var](https://docs.getdbt.com/reference/dbt-jinja-functions/var) method.
You can setup the vars to be configured using env_vars but I am not sure this is the case you want to support.

If we look at the [DbtRunner](https://github.com/elementary-data/elementary/blob/master/elementary/clients/dbt/dbt_runner.py) class we can see that we load the `dbt_env_vars` to the env and don't pass it as `vars` to the commands.

I think that a valid approach to this issue is to create an implementation that is similar to the `dbt_env_vars`, which will be used as default if specific `vars` were not passed to the `DbtRunner` commands.
@elongl  WDYT?

IDoneShaveIt: Hey @manulpatel, chiming in here 🙂
I just want to clarify and make sure you aware that this change will load the vars as env vars.
The meaning of this is that in order to access them you will have to use [env_var](https://docs.getdbt.com/reference/dbt-jinja-functions/env_var) method instead of the [var](https://docs.getdbt.com/reference/dbt-jinja-functions/var) method.
You can setup the vars to be configured using env_vars but I am not sure this is the case you want to support.

If we look at the [DbtRunner](https://github.com/elementary-data/elementary/blob/master/elementary/clients/dbt/dbt_runner.py) class we can see that we load the dbt_env_vars to the env and don't pass it as vars to the commands.

I think that a valid approach to this issue is to create an implementation that is similar to the dbt_env_vars, which will be used as default if specific vars were not passed to the DbtRunner commands.
@elongl WDYT?

elementary/config/config.py
```diff
@@ -67,7 +68,7 @@ def __init__(
         self.env = env
 
         # Additional env vars supplied to dbt invocations
-        self.dbt_env_vars = dict()
+        self.dbt_env_vars = dict() if dbt_vars is None else dbt_vars
```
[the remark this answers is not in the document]
manulpatel: Yes a cleaner approach!

elementary/monitor/cli.py
elongl: Yes, I agree, that is a good way to go here.
Adding `dbt_vars` alongside `dbt_env_vars`.

PR #854, opened 2023-05-08
Update anomaly-sensitivity.mdx

vishaalkk:
When adding sensitivity on a model level, it throws this keyword error:  `macro 'dbt_macro__test_volume_anomalies' takes no keyword argument 'anomaly_sensitivity'`

It should be `sensitivity` instead of an `anomaly_sensitivity`.


elongl:
Thanks for contributing this.

PR #898, opened 2023-05-26
Ele 801 support dbt 15 with dbtslimrunner

haritamar:


elementary/clients/dbt/slim_dbt_runner.py
```diff
@@ -23,15 +27,42 @@
...
 class ConfigArgs(BaseModel):
-    profiles_dir: Optional[str] = None
-    project_dir: str
+    project_dir: str = DEFAULT_PROJECT_DIR
+    profiles_dir: str = DEFAULT_PROFILES_DIR
+    profile: Optional[str] = None
     target: Optional[str] = None
     threads: Optional[int] = 1
-    vars: Optional[str] = DEFAULT_VARS
+    vars: Optional[Union[str, Dict[str, Any]]] = DEFAULT_VARS
```
elongl, line 65: Wouldn't it make more sense to always have that as a `dict` and do `json.dumps()` based on the version?

elementary/clients/dbt/slim_dbt_runner.py
```diff
@@ -23,15 +27,42 @@
 # Disable dbt tracking
 disable_tracking()
 
-DEFAULT_VARS = "{}"
+dbt_version = version.parse(dbt_version_string)
+
+DEFAULT_VARS: Union[str, Dict[str, Any]]
+if dbt_version >= version.parse("1.5.0"):
+    DEFAULT_VARS = {}
```
elongl, line 34: Review comment below.

elementary/clients/dbt/slim_dbt_runner.py
haritamar: This structure is passed to dbt, and vars need to be either a dict or a string depending on the version, so I must declare it as a union

elongl (approved):

PR #900, opened 2023-05-28
Ele 826 orchestration integration

NoyaArie:
Include in the report information about the last invocation for each resource.

### DIFF
--- a/elementary/monitor/api/invocations/invocations.py
+++ b/elementary/monitor/api/invocations/invocations.py
@@ -37,3 +37,10 @@ class InvocationsAPI(APIClient):
             )
         else:
             raise NotImplementedError
+
+    def get_invocations_by_ids(
+        self, invocations_ids: list[str]
+    ) -> DbtInvocationSchema:
+        return self.invocations_fetcher.get_invocations_by_ids(
+            macro_args=dict(ids=invocations_ids)
+        )
--- a/elementary/monitor/api/invocations/invocations.py
+++ b/elementary/monitor/api/invocations/invocations.py
@@ -40,9 +40,7 @@ class InvocationsAPI(APIClient):
         else:
             raise NotImplementedError
 
-    def get_invocations_by_ids(
-        self, invocations_ids: list[str]
-    ) -> DbtInvocationSchema:
+    def get_invocations_by_ids(self, invocations_ids: list[str]) -> DbtInvocationSchema:
         return self.invocations_fetcher.get_invocations_by_ids(
             macro_args=dict(ids=invocations_ids)
         )
--- a/elementary/monitor/api/invocations/invocations.py
+++ b/elementary/monitor/api/invocations/invocations.py
@@ -1,3 +1,5 @@
+from typing import Dict
+
 from elementary.clients.api.api_client import APIClient
 from elementary.clients.dbt.base_dbt_runner import BaseDbtRunner
 from elementary.monitor.fetchers.invocations.invocations import InvocationsFetcher
@@ -44,3 +46,6 @@ class InvocationsAPI(APIClient):
         return self.invocations_fetcher.get_invocations_by_ids(
             macro_args=dict(ids=invocations_ids)
         )
+
+    def get_resources_latest_invocation(self) -> Dict[str, str]:
+        return self.invocations_fetcher.get_resources_latest_invocation()
--- a/elementary/monitor/api/report/report.py
+++ b/elementary/monitor/api/report/report.py
@@ -3,6 +3,7 @@ from typing import Dict, List, Optional, Tuple
 
 from elementary.clients.api.api_client import APIClient
 from elementary.monitor.api.filters.filters import FiltersAPI
+from elementary.monitor.api.invocations.invocations import InvocationsAPI
 from elementary.monitor.api.lineage.lineage import LineageAPI
 from elementary.monitor.api.models.models import ModelsAPI
 from elementary.monitor.api.models.schema import (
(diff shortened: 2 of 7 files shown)


elementary/monitor/dbt_project/macros/get_resources_latest_invocation.sql
```diff
@@ -0,0 +1,22 @@
+{% macro get_resources_latest_invocation() %}
+  {% set dbt_run_results = ref('dbt_run_results') %}
+  {%- if elementary.relation_exists(dbt_run_results) -%}
```
IDoneShaveIt, line 3: I think that there is no need to check if this table exists.
If a user ran elementary dbt package he should have this table.
Plus I am not sure we want to do nothing if this table does not exist.

elementary/monitor/dbt_project/macros/get_resources_latest_invocation.sql
```diff
@@ -0,0 +1,22 @@
...
+    {% set get_resources_latest_invocation_query %}
+      with row_numbered_run_results as (
```
IDoneShaveIt, line 5: Lets give it a more meaningful name than row_numbered 🙂 
Something like `ordered_run_results` or something that explains what is the meaning of the row_number

elementary/monitor/dbt_project/macros/get_invocations_by_ids.sql
```diff
@@ -0,0 +1,19 @@
+{% macro get_invocations_by_ids(ids) %}
+  {% set database, schema = elementary.target_database(), target.schema %}
+  {% set invocations_relation = adapter.get_relation(database, schema, 'dbt_invocations') %}
+  {% if not invocations_relation %}
+    {% do elementary.edr_log('failed getting invocations relation') %}
+    {% do return(none) %}
+  {% endif %}
+
+  {% set get_invocations_query %}
+    select * from {{ invocations_relation }} where invocation_id in {{ elementary.strings_list_to_tuple(ids) }}
```
IDoneShaveIt, line 10: Lets spread `*` to make it easy to understand which data is returning from this query 🙂

elementary/monitor/dbt_project/macros/get_invocations_by_ids.sql
```diff
@@ -0,0 +1,19 @@
+{% macro get_invocations_by_ids(ids) %}
+  {% set database, schema = elementary.target_database(), target.schema %}
+  {% set invocations_relation = adapter.get_relation(database, schema, 'dbt_invocations') %}
+  {% if not invocations_relation %}
+    {% do elementary.edr_log('failed getting invocations relation') %}
```
IDoneShaveIt, line 5: Did you try to run this macro when there is no invocation_relation?
We used edr_log to be able to parse the results from our macro runs using the dbt runner.
I am not sure this will give you the wanted effect 🤔

elementary/monitor/dbt_project/macros/get_invocations_by_ids.sql
```diff
@@ -0,0 +1,19 @@
...
+    {% do return(none) %}
+  {% endif %}
+
+  {% set get_invocations_query %}
+    select * from {{ invocations_relation }} where invocation_id in {{ elementary.strings_list_to_tuple(ids) }}
+  {% endset %}
+  {% set result = elementary.run_query(get_invocations_query) %}
+  {% if not result %}
+    {% do elementary.edr_log('no invocations were found') %}
```
IDoneShaveIt, line 14: Same comment about the edr_log

elementary/monitor/api/invocations/invocations.py
```diff
@@ -37,3 +39,11 @@ def get_invocation_by_id(
             )
         else:
             raise NotImplementedError
+
+    def get_invocations_by_ids(self, invocations_ids: list[str]) -> DbtInvocationSchema:
```
IDoneShaveIt, line 43:
```suggestion
    def get_invocations_by_ids(self, invocations_ids: list[str]) -> List[DbtInvocationSchema]:
```

elementary/monitor/api/invocations/invocations.py
```diff
@@ -1,3 +1,5 @@
+from typing import Dict
```
IDoneShaveIt, line 1:
```suggestion
from typing import Dict, List
```

IDoneShaveIt (commented):
Most of the comments are small 🙂

elementary/monitor/api/report/report.py
```diff
@@ -102,6 +109,8 @@ def get_report_data(
                 model_runs_totals=serializable_model_runs_totals,
                 filters=serializable_filters,
                 lineage=serializable_lineage,
+                invocations=invocations,
+                resources_latest_invocation=resources_latest_invocation,
```
IDoneShaveIt, line 113: Do you need both of the new objects at the front?

IDoneShaveIt (approved):
LGTM!

elementary/monitor/api/invocations/invocations.py
JavierLopezT: Why? Using List is deprecated
https://peps.python.org/pep-0585/#implementation

```
list # typing.List

dict # typing.Dict

...

Importing those from typing is deprecated. Due to PEP 563 and the intention to minimize the runtime impact of typing, this deprecation will not generate DeprecationWarnings. Instead, type checkers may warn about such deprecated usage when the target version of the checked program is signalled to be Python 3.9 or newer. It's recommended to allow for those warnings to be silenced on a project-wide basis.

The deprecated functionality will be removed from the typing module in the first Python version released 5 years after the release of Python 3.9.0.
```

PR #905, opened 2023-05-29
Changed the `SidebarsAPI` to `GroupsAPI` that includes exposures.

elongl:
We want dbt exposures' tags and owners to show in the lineage API filters. Currently, the filters in the lineage are read from the `sidebars` object in the report's data. However, we don't want to add the exposures to the sidebars as it doesn't make sense. Therefore, we changed the `SidebarsAPI` to `GroupsAPI` in which each item in a group (tags, owners, etc.) also references its resource type. That way the report can exclude exposures from the sidebars but still include them in the lineage filter.


elementary/monitor/api/groups/groups.py
```diff
@@ -0,0 +1,106 @@
...
+
+            if artifact.owners:
+                for owner in artifact.owners:
+                    group[owner].append(self._get_group_item(artifact))
+            else:
+                group[NO_OWNERS_DEFAULT_TREE].append(self._get_group_item(artifact))
+        return dict(group)
+
+    @staticmethod
+    def _get_group_item(artifact: GROUPABLE_ARTIFACT) -> dict:
```
IDoneShaveIt, line 102: Maybe we want to create a schema for the grouped item as well?

elementary/monitor/api/groups/schema.py
```diff
@@ -0,0 +1,13 @@
+from typing import Dict, List, Union
+
+from pydantic import BaseModel
+
+DbtGroupSchema = Dict[str, Union[list, dict]]
+TagsGroupSchema = Dict[str, List[dict]]
+OwnersGroupSchema = Dict[str, List[dict]]
```
IDoneShaveIt, line 7: In continue to my prev comment about having a schema for the grouped item - it will be really nice to be used here so we would easily know what are the values of the group schemas contain 🙂

elementary/monitor/api/groups/schema.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class GroupsSchema(BaseModel):
+    dbt: DbtGroupSchema = dict()
```
IDoneShaveIt, line 11: Although it does work with mutable default values, it is not mentioned in pydantic documentation.
I am thinking that maybe it is safer to use `Field(default_factory=dict)` instead, but this is just might be me NIT picking 🙂

IDoneShaveIt (approved):

elementary/monitor/api/groups/groups.py
elongl: Yes!

elementary/monitor/api/groups/schema.py
elongl: [Seems fine](https://stackoverflow.com/a/63808835/8182628).

PR #912, opened 2023-06-03
Upgraded 'packaging' requirement.

elongl:


setup.py
```diff
@@ -49,7 +49,7 @@
         "pytest-parametrization>=2022.2.1",
         "pydantic<2.0",
         "networkx>=2.3,<3",
-        "packaging>=20.9,<23.0",
+        "packaging>=20.9,<=23.1",
```
IDoneShaveIt, line 52: Why do we want to use `<=` instead of just `<`

elongl: To allow this version as well as it does work.

IDoneShaveIt: So why not < 23.2?
I think that if this is how we work with the rest of our requirements we should do the same here 🙂

IDoneShaveIt (approved):

setup.py
elongl: `23.2` doesn't exist yet so I thought it might cause issues, but mostly, it doesn't really matter.
Also, we do have requirements that use `>=`.

PR #958, opened 2023-06-21
Pass warehouse type to FE.

elongl:


elementary/tracking/tracking_interface.py
```diff
@@ -27,46 +29,52 @@ def set_env(self, key: str, value):
...
     def _send_event(
+        self,
         distinct_id: str,
         event_name: str,
         properties: Optional[dict] = None,
-        groups: Optional[dict] = None,
     ) -> None:
         raise NotImplementedError
 
+    def register_warehouse_group(self, warehouse_info: WarehouseInfo) -> None:
```
IDoneShaveIt, line 46: I am not sure it should be here.
Not all of the tracking will need the register_warehouse_group  and this should be implemented only where needed (probably the anonymous tracking / outside of the tracking use)

elementary/monitor/data_monitoring/data_monitoring.py
```diff
@@ -36,7 +39,12 @@ def __init__(
         latest_invocation = self.get_latest_invocation()
         self.project_name = latest_invocation.get("project_name")
         dbt_pkg_version = latest_invocation.get("elementary_version")
+        self.warehouse_info = self._get_warehouse_info(
+            hash_id=isinstance(tracking, AnonymousTracking)
+        )
         if tracking:
+            if self.warehouse_info:
+                tracking.register_warehouse_group(self.warehouse_info)
```
IDoneShaveIt, line 47: You can just use `_register_group` here instead of having a specific method for that that is only used once

elementary/monitor/cli.py
```diff
@@ -288,6 +285,9 @@ def monitor(
             disable_samples=disable_samples,
             filter=select,
         )
+        anonymous_tracking.track_cli_start(
```
IDoneShaveIt, line 288: Don't forget to add a comment that explains the position of the tracking

IDoneShaveIt (approved):

PR #959, opened 2023-06-22
Ele 1194 test dwh connection edr

NoyaArie:


elementary/monitor/debug.py
```diff
@@ -17,6 +17,14 @@ def run(self) -> bool:
             self.config.profile_target,
         )
 
+        try:
+            dbt_runner.debug()
```
elongl, line 21: This is problematic because `dbt debug` requires `git` which is not necessarily required.

elongl (commented):
What problem does this PR solve?

elementary/monitor/debug.py
IDoneShaveIt: Do we know what is happening if we don't have git?
Does it still test the connection? If so, maybe we can extract only that part from the debug result?

NoyaArie:
> What problem does this PR solve?

1. When we failed to connect the DWH, the error is not informative.
2.  The first query we run in the report is getting the last invocation. We don't stop the execution when we were unable to connect the DWH at this point (only after). This is very confusing when we are trying to understand what went wrong.

PR #988, opened 2023-07-12
Ele 323 filter alerts by status

ellakz:


elementary/monitor/api/models/models.py
```diff
@@ -54,7 +54,7 @@ def get_models_runs(
                 ModelRunSchema(
                     id=model_run.invocation_id,
                     time_utc=model_run.generated_at,
-                    status=model_run.status,
+                    status=model_run.statuses,
```
haritamar, line 57: Any chance this was refactored by mistake?
I think it should still be `model_run.status`?

(and also the ones below)

ellakz: yep, thanks for catching this! fixed

haritamar (approved):

PR #990, opened 2023-07-12
MyPy support in OSS

haritamar:


setup.py
```diff
@@ -34,7 +34,7 @@
     "Snowflake, BigQuery, Redshift, data reliability, analytics engineering",
     long_description=README,
     install_requires=[
-        "click>=7.0,<9",
+        "click>=7.0,<8.1.4",
```
haritamar, line 37: There are typing bugs that were introduced specifically in click 8.1.4,
there's an active issue about it:
https://github.com/pallets/click/issues/2558

haritamar: Once they release a fix we can return the upper bound to be as it was before

elementary/monitor/api/totals_schema.py
```diff
@@ -0,0 +1,31 @@
+from pydantic import BaseModel
+
+
+class TotalsSchema(BaseModel):
```
haritamar, line 4: Was defined twice with exactly the same code (which caused wrong type references), so moved to a shared location

elementary/clients/dbt/slim_dbt_runner.py
```diff
@@ -126,23 +129,41 @@ def _load_config(self):
...
+            raise Exception("Config not loaded")
+
         register_adapter(self.config)
         self.adapter_name = self.config.credentials.type
-        self.adapter = get_adapter_class_by_name(self.adapter_name)(self.config)
-        self.adapter.connections.set_connection_name()
-        self.config.adapter = self.adapter
+        self.adapter = cast(
+            BaseAdapter, get_adapter_class_by_name(self.adapter_name)(self.config)
+        )
```
haritamar, line 139: Casts are needed here because dbt use `Protocol` classes as the return value types, and not the actual base classes
And these classes don't actually contain some of the methods we use here (which are internal so probably not a part of the official adapter protocol)

elementary/monitor/api/tests/tests.py
```diff
@@ -380,10 +393,15 @@ def _get_test_metadata_from_test_result_db_row(
...
         disable_samples: bool = False,
-    ) -> Union[DbtTestResultSchema, ElementaryTestResultSchema]:
-        test_results = None
+    ) -> Optional[Union[DbtTestResultSchema, ElementaryTestResultSchema]]:
+        test_results: Optional[Union[DbtTestResultSchema, ElementaryTestResultSchema]]
+
         sample_data = test_result_db_row.sample_data if not disable_samples else None
         if test_result_db_row.test_type == "dbt_test":
+            if sample_data is not None and not isinstance(sample_data, list):
+                # Sanity check, shouldn't happen
```
elongl, line 402: Don't really like this.
If it shouldn't happen there shouldn't be a check for it.
I think it means we're doing something wrong.

elementary/monitor/api/tests/tests.py
```diff
@@ -433,19 +459,25 @@ def _get_total_tests_results(
     ) -> Dict[str, TotalsSchema]:
         totals: Dict[str, TotalsSchema] = dict()
         for test in test_metadatas:
+            if not test.model_unique_id:
```
elongl, line 462: Not sure if there's something to do with it,
but this seems to be highly repeated, can / should we do it somewhere earlier in the flow?

elementary/monitor/api/tests/tests.py
```diff
@@ -433,19 +459,25 @@ def _get_total_tests_results(
...
+                continue
+
             self._update_test_results_totals(
                 totals_dict=totals,
                 model_unique_id=test.model_unique_id,
-                status=test.latest_run_status,
+                status=test.latest_run_status or "unknown",
```
elongl, line 468: How can `latest_run_status` be `None`?
Where does this `status` go to?

Reading the code seems like it won't do anything:
```py
    def add_total(self, status):
        total_adders = {
            "error": self._add_error,
            "warn": self._add_warning,
            "fail": self._add_failure,
            "pass": self._add_passed,
        }
        adder = total_adders.get(status)
        if adder:
            adder()
```

elementary/tracking/tracking_interface.py
```diff
@@ -93,15 +95,18 @@ def _group_identify(
                 },
             ),
         )
-        return response
+        if response.status_code != requests.codes.ok:
```
elongl, line 98: I think there's just `response.ok`.

.pre-commit-config.yaml
```diff
@@ -21,3 +21,12 @@ repos:
         name: Check for NO_COMMIT marker
         entry: bash -c "git diff --cached -U0 | (! grep NO_COMMIT)"
         language: system
+        require_serial: true
+
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: mypy --no-error-summary
```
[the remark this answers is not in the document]
haritamar: Just a bit spamming because it splits the `mypy` run to multiple processes and then outputs a line even when it succeeds, e.g:
image

If there are errors they will still be printed

elementary/cli/upgrade.py
```diff
@@ -10,6 +10,10 @@ def recommend_version_upgrade():
     try:
         latest_version = package.get_latest_package_version()
         current_version = package.get_package_version()
```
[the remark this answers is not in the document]
haritamar: hmm
I wasn't sure if `get_distribution` might fail
but I guess that's overly safe, so changing

elementary/clients/dbt/slim_dbt_runner.py
```diff
@@ -157,18 +178,24 @@ def _execute_macro(self, macro_name, **kwargs):
...
         macro_name: str,
         capture_output: bool = True,
-        macro_args: dict = dict(),
+        macro_args: Optional[dict] = None,
         log_errors: bool = True,
         vars: Optional[dict] = None,
         quiet: bool = False,
         **kwargs,
     ) -> list:
+        if self.profiles_dir is None:
```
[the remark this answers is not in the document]
haritamar: I recall this was broken in dbt 1.5
The profiles dir is parsed externally, so the config is already assumed to contain it.
I already changed back then the `__init__` function to make `profiles_dir` mandatory (only for the slim runner), but couldn't convince mypy here that it can't be None so added this check

elementary/monitor/alerts/group_of_alerts.py
```diff
@@ -237,7 +236,7 @@ def _get_model_error_block_body(self) -> str:
         return ""
 
     def _attention_required_blocks(self):
-        preview_blocks = [f"*{self._db}.{self._schema}.{self._model}*"]
```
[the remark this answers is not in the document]
haritamar: Yeah this code is pretty horrible
But `self._db`, `self._schema` and `self._model` are only defined in `GroupOfAlertsByTable` and not in `GroupOfAlerts`
So only used them there

elementary/monitor/api/tests/tests.py
```diff
@@ -148,17 +148,27 @@ def get_test_results(
...
+        tests_results: DefaultDict[str, List[TestResultSchema]] = defaultdict(list)
         for test_result_db_row in filtered_test_results_db_rows:
-            test_result = TestResultSchema(
-                metadata=self._get_test_metadata_from_test_result_db_row(
-                    test_result_db_row
-                ),
-                test_results=self._get_test_result_from_test_result_db_row(
-                    test_result_db_row, disable_samples=disable_samples
-                ),
+            if not test_result_db_row.model_unique_id:
```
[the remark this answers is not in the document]
haritamar: I don't think it really can happen, but this is one of the cases where I was afraid to make a field non optional :/
I can also raise an exception alternatively
Or do you think `cast` here is better? I saw also similar cases in your PR so wonder what is the best approach

elementary/monitor/api/tests/tests.py
haritamar: Same reason as above with all the "Optional"s

elementary/monitor/api/tests/tests.py
haritamar: Changing to cast

PR #991, opened 2023-07-13
Added better exception message for debugging

ofek1weiss:


elementary/exceptions/exceptions.py
```diff
@@ -82,6 +82,17 @@ def extract_detailed_dbt_command_args(command_args):
         except Exception as ex:
             logger.error(f"Failed to extract detailed dbt command args, error: {ex}")
 
+    def get_exception_message(self) -> Optional[str]:
+        error_output = self.proc_err.output.decode()
+        for line in reversed(error_output.strip().splitlines()):
```
elongl, line 87: Note that in other versions of dbt-core the `data` was renamed to `info`.
You can see an example of how we parse it [here](https://github.com/elementary-data/elementary/blob/4135a4530e481a49b9905e39a284a9adb39c72ec/elementary/clients/dbt/dbt_runner.py#L16) and [here](https://github.com/elementary-data/elementary/blob/4135a4530e481a49b9905e39a284a9adb39c72ec/elementary/clients/dbt/dbt_runner.py#L262).

I'd recommend we export this logic and share it between the two.

elementary/exceptions/exceptions.py
```diff
@@ -82,6 +82,17 @@ def extract_detailed_dbt_command_args(command_args):
         except Exception as ex:
             logger.error(f"Failed to extract detailed dbt command args, error: {ex}")
 
+    def get_exception_message(self) -> Optional[str]:
+        error_output = self.proc_err.output.decode()
```
elongl, line 86: `proc_err.output` can potentially be `None` if `check_output=False`, maybe worth adding a check here.
I think that we should raise an error in this case rather than return `None` as this is a coder problem.

elementary/monitor/debug.py
```diff
@@ -21,7 +21,7 @@ def run(self) -> bool:
             dbt_runner.run_operation("test_conn", quiet=True)
         except DbtCommandError as err:
             click.echo(
-                f"Could not connect to the Elementary db and schema. See details below\n\n{err}"
+                f"Could not connect to the Elementary db and schema. See details below\n\n{err.get_exception_message() or err}"
```
elongl, line 24: Thinking out loud, maybe we should just make `get_exception_message()` return `str(self)` if it fails rather than `None`?

elementary/clients/dbt/dbt_log.py
```diff
@@ -0,0 +1,38 @@
...
+@dataclass
+class DbtLog:
+    msg: Optional[str]
+    level: Optional[str]
+    exception: Optional[str]
+
+    @classmethod
+    def from_log_line(cls, log_line: str) -> 'DbtLog':
+        log = json.loads(log_line)
+        return cls(
```
elongl, line 20: Can we add a method for `get_json` that returns `log.get("info") or log.get("data") or {}`?
Afterwards I'd call it once and read the properties from it.

elementary/clients/dbt/dbt_log.py
```diff
@@ -0,0 +1,38 @@
...
+from elementary.utils.log import get_logger
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class DbtLog:
+    msg: Optional[str]
+    level: Optional[str]
+    exception: Optional[str]
```
elongl, line 15: I'd personally call it `err_msg`, but not critical.

elementary/clients/dbt/dbt_log.py
```diff
@@ -0,0 +1,38 @@
...
+        return cls(
+            msg=log.get("info", {}).get("msg") or log.get("data", {}).get("msg"),
+            level=log.get("info", {}).get("level") or log.get("level"),
+            exception=log.get("info", {}).get("exc") or log.get("data", {}).get("exc")
+        )
+    
+    def __str__(self) -> str:
+        ret = f"{self.level or 'unknown'}: {self.msg}"
+        if self.exception:
+            ret += f"\nException:\n{self.exception}"
```
elongl, line 29: I think this should say `Error:` as it's clearer to users than `Exception` which is a coding-term.

elementary/exceptions/exceptions.py
```diff
@@ -41,19 +42,30 @@ def __init__(
...
 
         # Command args sent to _run_command (without additional user-specific args it as such as projects / profiles
         # dir)
         self.proc_err = err
         self.base_command_args = base_command_args
         self.return_code = err.returncode
+        self.logs = logs
+    
+    @classmethod
+    def from_process_error(cls, err: subprocess.CalledProcessError, base_command_args: List[str], err_msg: Optional[str] = None) -> 'DbtCommandError':
```
elongl, line 63: Aren't the `base_command_args` and `err_msg` contained with the `err` object?
Also seems like this isn't used so maybe YAGNI.

By the way, any reason not to `-> DbtCommandError`?

elementary/exceptions/exceptions.py
```diff
@@ -41,19 +42,30 @@ def __init__(
...
+        if err.output is None:
+            return cls(err, base_command_args, err_msg, None)
```
elongl, line 65: Can you be explicit and pass `logs=None`?

elementary/clients/dbt/dbt_log.py
ofek1weiss: sadly it wont always work (for example in `level`)

elementary/clients/dbt/dbt_log.py
ofek1weiss: i like it better in the code as `exception` seeing as its name in dbt is `exc`, but i will change the `__str__` usage to Error

elementary/clients/dbt/dbt_log.py
ofek1weiss: On it

elementary/clients/dbt/dbt_log.py
```diff
@@ -0,0 +1,38 @@
...
+        return ret
```
IDoneShaveIt, line 30: Use an explicit var instead of "ret"

elementary/exceptions/exceptions.py
ofek1weiss: `base_command_args` and `err.args` are not the same :(
as for `err_msg`, its a way for the programmer to give a more specific error message, by default it is the error logs

elementary/exceptions/exceptions.py
ofek1weiss: Sure

elongl (approved):

PR #1031, opened 2023-07-26
Fix docker build pip version (causing package installation issues)

noaKurman:
I can just run `--upgrade pip` but I feel like setting it to a specific version is a better practice

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update \
     /var/tmp/*
 
 COPY . /app
+RUN pip install --upgrade pip
 RUN pip install --no-cache-dir "/app[all]"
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apt-get update \
     /var/tmp/*
 
 COPY . /app
+RUN pip install --upgrade pip==23.2.1
```
elongl, line 21: From all the Dockerfiles I've ever seen, they just upgrade and not specify a specific version.
I agree that it's the practice with libraries, but in `pip` I think latest is preferable.
Anyhow, approving, up to you.

elongl (approved):

PR #1047, opened 2023-08-02
Ele 638 azure blob storage and elementary data cli

ellakz:


elementary/clients/azure/client.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def send_report(
+        self, local_html_file_path: str, remote_bucket_file_path: Optional[str] = None
+    ) -> Tuple[bool, Optional[str]]:
+        report_filename = (
+            bucket_path.basename(remote_bucket_file_path)
+            if remote_bucket_file_path
+            else path.basename(local_html_file_path)
+        )
+        blob_service_client = BlobServiceClient.from_connection_string(
```
IDoneShaveIt, line 34: I think that it will be nice if the blob_service_client will be initialized at the __init__

elementary/clients/azure/client.py
```diff
@@ -0,0 +1,56 @@
...
+            self.config.azure_connection_string
+        )
+        self.client = blob_service_client.get_blob_client(
```
IDoneShaveIt, line 37: This is specific to a container and a blob, so it should not be set as the AzureClient self.client (each blob will have his own and it is not "general")

elementary/clients/azure/client.py
IDoneShaveIt: It looks like this will be the class global client that will be used to connect to the blobs

IDoneShaveIt (approved):
Looking good!
Approved with some small comments!

daandes:
Thanks for this, any idea on when is this being released?

PR #1051, opened 2023-08-02
alerts bugfix - ensure channel in alerts_config works in grouped mode

haritamar:


elementary/monitor/alerts/test.py
```diff
@@ -34,11 +34,14 @@ def __init__(
         test_unique_id: str,
         test_name: str,
         test_created_at: Optional[str] = None,
-        test_meta: Optional[str] = None,
+        test_meta: Optional[Union[str, dict]] = None,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
-        self.test_meta = try_load_json(test_meta) or {}
+        if isinstance(test_meta, dict):
```
IDoneShaveIt, line 41: NIT: maybe we can create another util that is called try_load_dict_json (because json could be a list as well) which will do this exact if statement?

IDoneShaveIt (approved):
LGTM with one NIT

PR #1056, opened 2023-08-03
Ele 1466 dwh connection docs

NoyaArie:


Maayan-s:
Added your changes and merged on another branch

PR #1069, opened 2023-08-08
Bump pydantic dependancy

mhindery:
We are trying to use this in a project where the restriction on Pydantic makes it incompatible with other packages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         "alive-progress<=2.3.1",
         "slack-sdk>=3.20.1,<4.0.0",
         "pytest-parametrization>=2022.2.1",
-        "pydantic<2.0",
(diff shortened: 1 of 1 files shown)


elongl:
Hi @mhindery, may I ask why don't you use a separate [venv](https://docs.python.org/3/library/venv.html) for the other application?

crabio:
Hi!
I have the same issue with Airflow.
Airflow 2.8.1 requires `pydantic>=2.3.0`, but `elementary-data` requires `pydantic<2.0`

May we merge it? What's the case to block this bump?

ellakz:
Hi, just merged a different solution to this issue so closing this PR, edr should support both pydantic 1 and 2 when we release the next version (0.14.1)

PR #1082, opened 2023-08-15
Notify tests fails to slack

noaKurman:


noaKurman:
General comment - I know there's code duplication here, which I might could have exported to a separated workflow, but then I'd need to use "wait for" which is much slower, so for now this feels good enough

.github/workflows/test-all-warehouses.yml
```diff
@@ -44,3 +44,35 @@ jobs:
       dbt-version: ${{ inputs.dbt-version }}
       generate-data: ${{ inputs.generate-data || false }}
     secrets: inherit
+
+  notify_failures:
+    name: Notify Slack
+    needs: test
+    if: |
+      always() &&
+      contains(needs.test.result, 'failure') &&
```
elongl, line 53: Why not use [`failure()`](https://docs.github.com/en/actions/learn-github-actions/expressions#failure)?

.github/workflows/test-all-warehouses.yml
```diff
@@ -44,3 +44,35 @@ jobs:
       dbt-version: ${{ inputs.dbt-version }}
       generate-data: ${{ inputs.generate-data || false }}
     secrets: inherit
+
+  notify_failures:
```
elongl, line 48: Not sure I understand [this comment](https://github.com/elementary-data/elementary/pull/1082#issuecomment-1680713605). Why not reference the job in the different workflow? like we do here with `test-warehouse.yml`

.github/workflows/test-all-warehouses.yml
noaKurman: It's step vs. jobs, not the same 
Not sure it'd work here

.github/workflows/test-all-warehouses.yml
noaKurman: I need to pass the jobs status to the uses and as far as I know its not possible
I can upload an artifact but at this point it might be an overhead for just those two things

elongl (approved):

.github/workflows/notify_slack.yml
```diff
@@ -0,0 +1,39 @@
+name: Test all warehouse platforms
```
elongl, line 1: This needs to be `Notify Slack`.

.github/workflows/notify_slack.yml
```diff
@@ -0,0 +1,39 @@
...
+        required: false
+      workflow_name:
+        type: string
+        required: true
+
+jobs:
+  notify:
+    name: Notify Slack Failure
+    runs-on: ubuntu-latest
+    if: ${{ inputs.result == 'failure' }}
```
elongl, line 19: The caller should choose when to call this job.
If someone called `Notify Slack` it should run.

.github/workflows/notify_slack.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  workflow_call:
+    inputs:
+      result:
+        type: string
+        required: true
+      run_id:
+        type: string
+        required: false
+      workflow_name:
```
elongl, line 11: This is not used anywhere, I'm assuming it needs to be in the Slack message.

.github/workflows/test-github-action.yml
```diff
@@ -101,3 +101,17 @@ jobs:
...
+    if: |
+      always() &&
+      contains(needs.test.result, 'failure') &&
+      ! cancelled() &&
+      ${{ github.event_name == 'schedule' }}
+    uses: ./.github/workflows/notify_slack.yml
+    with:
+      result: "failure"
+      run_id: ${{ github.run_id }}
+      workflow_name: "Test elementary GitHub action"
```
elongl, line 117:
```suggestion
      workflow_name: "Test Elementary GitHub action"
```

Forgive my OCD 😅

.github/workflows/test-all-warehouses.yml
```diff
@@ -45,3 +45,17 @@ jobs:
       dbt-version: ${{ matrix.dbt-version }}
       generate-data: ${{ inputs.generate-data || false }}
     secrets: inherit
+
+  notify_failures:
+    name: Notify Slack
+    needs: test
+    if: |
+      always() &&
+      contains(needs.test.result, 'failure') &&
```
elongl, line 54: Is there a reason not to do string comparison?

```suggestion
      needs.test.result == 'failure' &&
```

.github/workflows/notify_slack.yml
noaKurman: Woops

.github/workflows/notify_slack.yml
noaKurman: Woops 2

.github/workflows/notify_slack.yml
noaKurman: I'm coloring the message, so if we'll want to add a "success" message we'll add another job with a different if statement. For now doesn't feel like a big deal in the implementation

.github/workflows/test-all-warehouses.yml
noaKurman: Actually not sure it'd work, I thinkkk there's a reason why this is how its mentioned to do in the docs 
Anyway would rather leave it like that instead of taking the chance, it works :)

.github/workflows/notify_slack.yml
elongl: If you're not going to change it, then I think we should rename this workflow and step to `notify_slack_failure` and remove the `if`.

noaKurman: But I might add a success job. thats what I'm saying

PR #1085, opened 2023-08-15
Not passing a long CLI arg to macro.

elongl:
Resolves [this](https://elementary-community.slack.com/archives/C02CTC89LAX/p1689914999477479).


elementary/monitor/api/report/report.py
```diff
@@ -90,12 +90,8 @@ def get_report_data(
...
             serializable_lineage = lineage.dict()
 
-            resources_latest_invocation = (
-                invocations_api.get_resources_latest_invocation()
-            )
-            invocations = invocations_api.get_invocations_by_ids(
-                invocations_ids=list(set(resources_latest_invocation.values()))
-            )
+            models_latest_invocation = invocations_api.get_models_latest_invocation()
+            invocations = invocations_api.get_models_latest_invocations_data()
```
haritamar, line 94: Only thing I'm slightly worried about here is a race between the two calls -  if a dbt pipeline is running at the same time it's possible the two calls won't return the same IDs.
(maybe feels unlikely, but I've seen such a race in the past in the CLI :) )

I understood your point about keeping the IDs macro, but wondering here since we're running both of them if we can just compute `models_latest_invocation` from `invocations`.
WDYT?

haritamar: Isn't it possible that you'll have an old invocation in `models_latest_invocation` and a new invocation in `invocations`?

haritamar (approved):

PR #1088, opened 2023-08-17
Run CI on both lowest and highest versions of dbt.

elongl:


.github/workflows/test-all-warehouses.yml
```diff
@@ -27,6 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        dbt-version: ["1.3.0", null]
```
elongl, line 30: Schema names will conflict.

.github/workflows/test-warehouse.yml
```diff
@@ -130,6 +122,15 @@ jobs:
...
           pip install ".[${{ (inputs.warehouse-type == 'databricks_catalog' && 'databricks') || inputs.warehouse-type }}]"
 
+      - name: Write dbt profiles
+        env:
+          PROFILES_YML: ${{ secrets.CI_PROFILES_YML }}
+        run: |
+          mkdir -p ~/.dbt
+          DBT_VERSION=$(pip show dbt-core | grep -i version | awk '{print $2}' | sed 's/\.//g')
+          UNDERSCORED_REF_NAME=$(echo "dbt_$DBT_VERSION_$BRANCH_NAME" | head -c 32 | sed "s/-/_/g")
+          echo "$PROFILES_YML" | base64 -d | sed "s/<SCHEMA_NAME>/py_$UNDERSCORED_REF_NAME/g" > ~/.dbt/profiles.yml
```
elongl, line 132: Added the dbt version so that schemas won't conflict when running concurrently.

.github/workflows/test-warehouse.yml
```diff
@@ -130,6 +123,15 @@ jobs:
           pip install -r dev-requirements.txt
           pip install ".[${{ (inputs.warehouse-type == 'databricks_catalog' && 'databricks') || inputs.warehouse-type }}]"
 
+      - name: Write dbt profiles
+        env:
+          PROFILES_YML: ${{ secrets.CI_PROFILES_YML }}
+        run: |
+          mkdir -p ~/.dbt
+          DBT_VERSION=$(pip show dbt-core | grep -i version | awk '{print $2}' | sed 's/\.//g')
```
elongl, line 131: Should this be `inputs.dbt-version` because this is what defines the concurrency group?

noaKurman: Not sure I understood the quesetion

.github/workflows/test-all-warehouses.yml
elongl, line 30: This is a bit annoying when using `workflow_dispatch` because it'll create jobs that'll immediately fail like [this](https://github.com/elementary-data/elementary/actions/runs/5923924227).

.github/workflows/test-warehouse.yml
noaKurman: Oh if you're asking if you should use the version from the inputs that I think you should

noaKurman (approved):
BTW do we have any limits on the warehouses? I wonder if it'll be slower when two pipelines will run at the same time

.github/workflows/test-all-warehouses.yml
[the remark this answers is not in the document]
elongl: Not quite, we're supporting 1.3.0 onwards.

.github/workflows/test-all-warehouses.yml
```diff
@@ -41,6 +42,6 @@ jobs:
       warehouse-type: ${{ matrix.warehouse-type }}
       elementary-ref: ${{ inputs.elementary-ref }}
       dbt-data-reliability-ref: ${{ inputs.dbt-data-reliability-ref }}
-      dbt-version: ${{ inputs.dbt-version }}
+      dbt-version: ${{ inputs.dbt-version || matrix.dbt-version }}
```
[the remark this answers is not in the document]
elongl: Yeah, it'll just assign this `null`.

.github/workflows/test-warehouse.yml
elongl: It's just a bit annoying because half of the time it'll be empty so I prefer to remain consistent and always have the dbt version in the schema name.

PR #1122, opened 2023-08-28
Refactor context manager nesting

kovaacs:
It's best practice not to keep context managers open if they're not needed. Reordering the code makes things much clearer.


elongl (approved):
Thanks for the contribution!

PR #1126, opened 2023-08-29
Ele 1625 add link to test runs in alerts

NoyaArie:


elementary/monitor/alerts/test.py
```diff
@@ -318,6 +332,17 @@ def to_slack(self, is_slack_workflow: bool = False) -> SlackMessageSchema:
                 ),
             )
 
+        test_runs_report_link = get_test_runs_report_link(
```
ellakz, line 335: maybe i'd rename that too just so that it's clear that it's meant for the alerts and not the report. something like `get_test_runs_url_for_alert`

elementary/monitor/alerts/report_link_utils.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional
+
+from elementary.monitor.alerts.schema.alert import ReportLinkData
+
+LINK_TEXT = "Runs history"
+
+
+def _get_formatted_report_url(report_url: str) -> str:
+    return report_url[:-1] if report_url.endswith("/") else f"{report_url}"
```
ellakz, line 9: I think you can achieve the same thing using `report_url.strip('/')`

ellakz (approved):
LGTM :) 
Left some comments mainly on naming but nothing blocking

elementary/monitor/alerts/test.py
NoyaArie: This file is under `monitor/alerts`, do you think `for_alert` should still be added to the name?

PR #1137, opened 2023-09-04
Alert links updates

NoyaArie:
- change copy in case of a model run link
- remove link in case of a model error in grouped alerts

### DIFF
--- a/elementary/monitor/alerts/report_link_utils.py
+++ b/elementary/monitor/alerts/report_link_utils.py
@@ -3,13 +3,13 @@ from typing import Optional
 
 from elementary.monitor.alerts.schema.alert import ReportLinkData
 
-TEST_LINK_TEXT = "View test runs"
-MODEL_LINK_TEXT = "View model runs"
+TEST_RUNS_LINK_TEXT = "View test runs"
+MODEL_RUNS_LINK_TEXT = "View model runs"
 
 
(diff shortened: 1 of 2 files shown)


elementary/monitor/alerts/report_link_utils.py
```diff
@@ -1,37 +1,48 @@
...
 
 from elementary.monitor.alerts.schema.alert import ReportLinkData
 
-LINK_TEXT = "View test runs"
+TEST_LINK_TEXT = "View test runs"
+MODEL_LINK_TEXT = "View model runs"
+
+
+class ReportPath(Enum):
+    TEST = "test-runs"
```
elongl, line 11: I think those should have their full names (`TEST_RUNS`, `MODEL_RUNS`) so that it wouldn't confuse with the test results screen.

elementary/monitor/alerts/report_link_utils.py
```diff
@@ -1,37 +1,48 @@
+from enum import Enum
 from typing import Optional
 
 from elementary.monitor.alerts.schema.alert import ReportLinkData
 
-LINK_TEXT = "View test runs"
+TEST_LINK_TEXT = "View test runs"
```
elongl, line 6: Same as comment below.

elongl (approved):

PR #1154, opened 2023-09-11
Add docs a for exposure generic tests

erikzaadi:
Depends on https://github.com/elementary-data/dbt-data-reliability/pull/530

Added docs for the new exposure validation tests:

image
image
image


docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: "Add exposure validation tests"
+---
+
+After you [install the dbt package](/quickstart#install-the-dbt-package), you can add Elementary exposure validation tests.
+
+## Exposure validation dbt tests
+
+Elementary dbt package includes **exposure validation tests, implemented as [dbt tests](https://docs.getdbt.com/docs/building-a-dbt-project/tests)**.
+These tests can detect missing columns and column data types for defined exposures.
```
Maayan-s, line 10: @erikzaadi 
This is a technical description of the test. 
I think it's better to describe here the use case it solves. 
"These tests will fail if there are column changes that break downstream exposures, such as BI dashboards."

Also, it's a good place to say that Elementary Cloud automatically generates exposures, maps the lineage, and adds the test to the models that the exposures depend on.

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,94 @@
...
+    tests:
+      - elementary.volume_anomalies:
+          tags: ["table_anomalies"]
+          timestamp_column: "order_date"
+      - exposure_schema_validity:
+          tags: [elementary]
+
+```
+
+We recommend adding a tag to the tests so you could execute these in a dedicated run using the selection parameter `--select tag:elementary`.
```
ellakz, line 83: do we have this recommendation in all the test types?

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,94 @@
...
+Upon running the tests, if any issues will arise, you'll see your tests fail and in the test result query you'll be able to see the reasons:
+
+```sql
+SELECT 'customers' as exposure, 'https://your.bi.tool/dashboards/1' as url, 'different data type for the column customer_id numeric vs numeric' as error
+UNION ALL SELECT 'customers' as exposure, 'https://your.bi.tool/dashboards/1' as url, 'order_id column missing in the model' as error
+```
+
+## What does it mean when a test fails?
+
+When a test fail, it means that an exposure is potentially broken, open the your BI tool to validate, and if needed update the `exposures.yml` or your model's schema.
```
ellakz, line 94:
```suggestion
When a test fails, it means that an exposure is potentially broken due to a missing or wrongly-typed column. open the your BI tool to validate, and if you make any changes to your dashboards be sure to update the `exposures.yml` and your model's schema accordingly.
```

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,94 @@
...
+  owner:
+    name: Callum McData
+    email: data@jaffleshop.com
+  meta:
+    columns:
+      - name: "order_id"
+        data_type: "numeric"
+```
+
+Then in your module schema, add the elementary exposure tests:
```
ellakz, line 65: I think it's important to point out that the test needs to be set up for every model used in the exposure.
Also you might wanna explain a little on how the test works, mainly to communicate why no other setup is needed except this test :)

docs/guides/add-exposure-tests.mdx
erikzaadi: saw it in https://github.com/elementary-data/elementary/blob/ele-1703-create-generic-test-for-exposure-stability/docs/guides/add-elementary-tests.mdx#L234

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "Add exposure validation tests"
+---
+
+After you [install the dbt package](/quickstart#install-the-dbt-package), you can add Elementary exposure validation tests.
+
+## Exposure validation dbt tests
+
+Elementary dbt package includes **exposure validation tests, implemented as [dbt tests](https://docs.getdbt.com/docs/building-a-dbt-project/tests)**.
+These tests can test that there are column changes that break downstream exposures, such as BI dashboards.
```
ellakz, line 10:
```suggestion
These tests can detect changes in your models' columns that break downstream exposures, such as BI dashboards.
```

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,157 @@
...
+    name: Callum McData
+    email: data@jaffleshop.com
+  meta:
+    referenced_columns:
+      - name: "order_id"
+        data_type: "numeric"
+        source: ref('returned_orders')
+      - name: "customer_id"
+        data_type: "numeric"
+        source: ref('orders')
```
ellakz, line 122: Not sure it's possible, but it will be nice if we could do something like:
referenced_columns:
  ref('orders'):
    - name: "order_id"
      data_type: "numeric"
    - name: "customer_id"
      data_type: "numeric"
  ref('customers'):
    - ...

docs/guides/add-exposure-tests.mdx
```diff
@@ -0,0 +1,157 @@
...
+          tags: ["table_anomalies"]
+          timestamp_column: "order_date"
+      - elementary.exposure_schema_validity:
+          tags: [elementary]
+
+```
+
+We recommend adding a tag to the tests so you could execute these in a dedicated run using the selection parameter `--select tag:elementary`.
+
+Upon running the tests, if any issues arise, you'll see your tests fail and in the test result query you'll be able to see the reasons:
```
ellakz, line 148:
```suggestion
Upon running the tests, if breaking changes are detected in the model, the test will fail and in the test result query you'll be able to see the reasons:
```

docs/guides/add-exposure-tests.mdx
erikzaadi: not sure, and it'll make the "simple" workflow (AKA no need to set the source property since there's only one exposure dependency) obsolute, which IMHO is a shame, it's prolly our 80 vs 20 scenario

erikzaadi: Pushed, only difference is `node` instead of `model_name` as per our conversation and @haritamar 's comment on the main PR

ellakz (approved):

PR #1168, opened 2023-09-19
Updated prioritization of timestamp columns.

elongl:


elementary/monitor/dbt_project/models/tests_recommendation/table_timestamp_columns.sql
```diff
@@ -76,6 +81,18 @@ with
...
     ),
 
+    -- Users can provide the timestamp columns for their models,
+    -- if provided, we assign a confidence score of 0 (certain).
+    model_provided_timestamp_columns as (
+        select
+            lower(database_name) as database_name,
+            lower(schema_name) as schema_name,
+            lower(name) as table_name,
+            lower(partition_by_field) as column_name
```
elongl, line 91: We need to make sure that this field is a `date/timestamp` since you can also partition by an integer.

elementary/monitor/dbt_project/models/tests_recommendation/table_timestamp_columns.sql
```diff
@@ -1,35 +1,35 @@
+{# Prioritization: 1. insertion time, 2. update time. #}
 {% set timestamp_column_names = [
+    "timestamp",
```
haritamar, line 3: I think this is also likely an event / data timestamp

elementary/monitor/dbt_project/models/tests_recommendation/table_timestamp_columns.sql
```diff
@@ -76,6 +76,18 @@ with
...
+    -- Users can provide the timestamp columns for their models,
+    -- if provided, we assign a confidence score of 0 (certain).
+    model_provided_timestamp_columns as (
+        select
+            lower(database_name) as database_name,
+            lower(schema_name) as schema_name,
+            lower(name) as table_name,
+            lower(bigquery_partition_by_field) as column_name
+        from {{ ref("elementary", "dbt_models") }}
+        where bigquery_partition_by_field is not null
```
haritamar, line 88: Note that BQ partitions may be on columns that are not timestamps (e.g. [here](https://cloud.google.com/bigquery/docs/partitioned-tables#integer_range))
so I think checking the type is also important.
(related to what I wrote in the dbt PR - I think it's good to save the entire `partition_by` dict)

elementary/monitor/dbt_project/models/tests_recommendation/table_timestamp_columns.sql
elongl: Deleted.

haritamar (approved):

PR #1169, opened 2023-09-19
ElE-1668: Docs

dapollak:


docs/guides/anomaly-detection-configuration/ignore_small_changes.mdx
```diff
@@ -0,0 +1,62 @@
...
+---
+
+```
+ignore_small_changes:
+  spike_failure_percent_threshold: [int]
+  drop_failure_percent_threshold: [int]
+```
+
+If defined, an anomaly test will fail only if all the following conditions hold:
+- The z-score of the metric within the detection period is anomoulous
```
ellakz, line 13: it's not accurate.
the test will fail if 1 AND (2 OR 3)

docs/guides/anomaly-detection-configuration/ignore_small_changes.mdx
```diff
@@ -0,0 +1,62 @@
...
+- The metric within the detection period is higher than `spike_failure_percent_threshold` percentages of the mean value in the training period, if defined.
+- The metric within the detection period is lower than `drop_failure_percent_threshold` percentages of the mean value in the training period, if defined
+
+Those settings can help to deal with
```
ellakz, line 17: unfinished sentence?
Also I think it's a good idea to add a graph here, you can add one from Adam's Figma

docs/guides/anomaly-detection-configuration/fail_on_zero.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: "fail_on_zero"
+sidebarTitle: "fail_on_zero"
+---
+
+`fail_on_zero: true/false`
+
+Elementary anomaly detection tests will fail whether there is any zero metric value within the detection period.
```
ellakz, line 8:
```suggestion
Elementary anomaly detection tests will fail if there is a zero metric value within the detection period.
```

docs/guides/anomaly-detection-configuration/fail_on_zero.mdx
```diff
@@ -0,0 +1,37 @@
...
+If undefined, default is false.
+
```
ellakz, line 10: maybe we can add a graph? Do we have something on Figma?

docs/guides/anomaly-detection-configuration/fail_on_zero.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: "fail_on_zero"
```
ellakz, line 2: This setting should only be applicable to volume anomalies, and needs to also appear in the volume anomalies parameter list [here](https://docs.elementary-data.com/guides/anomaly-detection-tests/volume-anomalies#test-configuration)

docs/guides/anomaly-detection-configuration/ignore_small_changes.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: "ignore_small_changes"
+sidebarTitle: "ignore_small_changes"
```
ellakz, line 3: This should also be added to the param list of all the tests (like [here](https://docs.elementary-data.com/guides/anomaly-detection-tests/volume-anomalies#test-configuration))

ellakz (changes requested):

docs/guides/anomaly-detection-configuration/ignore_small_changes.mdx
dapollak: @ellakz 
Completed the sentence 😆 
Not sure there is a graph that reflects exactly those params, because with our last call it looks the same with just wider range

docs/guides/anomaly-detection-configuration/fail_on_zero.mdx
dapollak: [https://github.com/elementary-data/elementary/pull/1169#discussion_r1336915277](Same)

ellakz (approved):

PR #1188, opened 2023-10-01
Ele 1804 source freshness result description calculate

NoyaArie:


elementary/utils/time.py
```diff
@@ -70,6 +72,16 @@ def convert_datetime_utc_str_to_timezone_str(
         return isoformat_datetime
 
 
+def convert_datetime_utc_str_to_timezone_datatime(
```
IDoneShaveIt, line 75: We already have the following 2 methods:
`convert_utc_iso_format_to_datetime`
`convert_utc_time_to_timezone`
You can use both or at least call those 2 inside this method instead of rewriting the logic that might be change in one place but be left the same on the other 🙂

elementary/utils/time.py
```diff
@@ -83,3 +95,28 @@ def convert_partial_iso_format_to_full_iso_format(partial_iso_format_time: str)
             f'Failed to covert time string: "{partial_iso_format_time}" to ISO format'
         )
         return partial_iso_format_time
+
+
+def get_formatted_timedelta(time_ago_in_s: float) -> str:
```
IDoneShaveIt, line 100: Just for later use - does it work with int as well?
If it does I think we should add it in the typing as mypy can start scream on other users that will use it due to it 🤔

tests/mocks/fetchers/alerts_fetcher_mock.py
```diff
@@ -317,7 +317,7 @@ def query_pending_source_freshness_alerts(self, *args, **kwargs):
                 detected_at=CURRENT_TIMESTAMP_UTC,
                 snapshotted_at="2022-10-11 10:00:00",
                 max_loaded_at="2022-10-11 10:00:00",
-                max_loaded_at_time_ago_in_s="123123",
+                max_loaded_at_time_ago_in_s=123123,
```
IDoneShaveIt, line 320: I see that we add those fields but do we test that we get the wanted result from them?

elementary/utils/time.py
IDoneShaveIt: Can we add tests for that?

elementary/monitor/alerts/source_freshness.py
```diff
@@ -31,17 +37,39 @@ def __init__(
...
         )
+
         self.max_loaded_at = (
-            convert_datetime_utc_str_to_timezone_str(max_loaded_at, self.timezone)
+            convert_datetime_utc_str_to_timezone_datatime(max_loaded_at, self.timezone)
             if max_loaded_at
             else None
         )
+        self.max_loaded_at_str = (
+            self.max_loaded_at.strftime(DATETIME_FORMAT) if self.max_loaded_at else None
```
IDoneShaveIt, line 52: Lets extend the util `convert_datetime_utc_str_to_timezone_str` to support `with_time_zone` which by default is false to maintain backwards, instead of using the strftime all the time.

If `with_time_zone` is false - use `DATETIME_FORMAT`
else - use `DATETIME_WITH_TIMEZONE_FORMAT`

elementary/utils/time.py
IDoneShaveIt: If we do keep this method that calls the other methods, can you add tests for it please 🙂

IDoneShaveIt (commented):
Looking good overall.
Few small comments + lets add some tests to the time utils you touched (this is an important part of the system and we need to start make sure those methods are stable!)

IDoneShaveIt (approved):
LGTM

PR #1190, opened 2023-10-01
Use enriched exposures view

erikzaadi:
Use `enriched_exposures` view

(Consolidates `dbt_exposures` and `enriched_exposures`)


NoyaArie (approved):
image

PR #1193, opened 2023-10-02
Ele-1834 Backwards compatibility for source freshness new fields

NoyaArie:


elementary/monitor/dbt_project/macros/alerts/get_source_freshness_alerts.sql
```diff
@@ -46,7 +55,12 @@
                 sources.freshness_error_after,
                 sources.freshness_warn_after,
                 sources.freshness_filter,
-                sources.freshness_description
+                {# backwards compatibility #}
+                {% if freshness_description_column_exists %}
+                    sources.freshness_description
+                {% else %}
+                    "dbt source freshness validates if the data in a table is not updated by calculating if the time elapsed between the test execution to the latest record is above an acceptable SLA threshold." as freshness_description
```
elongl, line 62: This needs to be in single quotes.

elementary/monitor/dbt_project/macros/alerts/get_source_freshness_alerts.sql
```diff
@@ -1,5 +1,11 @@
 {% macro get_pending_source_freshness_alerts(days_back) %}
     -- depends_on: {{ ref('alerts_source_freshness') }}
+    {% set alerts_source_freshness_relation = ref('alerts_source_freshness') %}
+    {% set error_after_column_exists = elementary.column_exists_in_relation(alerts_source_freshness_relation, 'error_after') %}
```
elongl, line 4: Can you get the columns once with `adapter.get_columns_in_relation` and then check for the required columns?
We're re-running it each time in `column_exists_in_relation`.

Here's how I'd do it.
At `column_exists_in_relation`, I'd add `columns=none`, and they load them if they're `none`, otherwise use `columns`.

elongl (approved):

PR #1218, opened 2023-10-08
Move `enriched_exposures` and `elementary_exposures` from dbt package

erikzaadi:
Depends on https://github.com/elementary-data/dbt-data-reliability/pull/583

### DIFF
--- a/elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
+++ b/elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
@@ -1,0 +1,19 @@
+{{
+config(
+  materialized = 'view',
+  bind=False,
+  unique_key='unique_id',
+)
+}}
+
+with dbt_exposures as (
+select * from {{ ref('dbt_exposures') }}
+),
+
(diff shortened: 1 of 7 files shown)


haritamar (approved):

elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
```diff
@@ -0,0 +1,19 @@
...
+with dbt_exposures as (
+select * from {{ ref('dbt_exposures') }}
+),
+
+elementary_exposures as (
+select * from {{ ref('elementary_exposures') }}
+)
+
+{# Union without duplicates where elementary_exposures has prio #}
+select * from dbt_exposures where not exists (select 1 from elementary_exposures where dbt_exposures.unique_id = elementary_exposures.unique_id)
```
haritamar, line 18: I wonder if this will be more efficient:
```sql
where dbt_exposures.unique_id not in (select elementary_exposures.unique_id from elementary_exposures)
```

because the subquery doesn't use any outer fields, so I think it can be cached.
WDYT?

PR #1237, opened 2023-10-23
Ele 994 detection delay docs

ellakz:


docs/guides/anomaly-detection-tests/all-columns-anomalies.mdx
```diff
@@ -36,6 +36,9 @@ No mandatory configuration, however it is highly recommended to configure a `tim
       &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<a href="/guides/anomaly-detection-configuration/time-bucket"><font color="#CD7D55">period: [hour | day | week | month]</font></a>
```
erikzaadi, line 36: Are all the whitespaces needed here? Perhaps there's a helper css class that adds margin instead?
In addition, isn't this the same link block that's used later? Wonder if there's a snippet or something to DRY it up.

erikzaadi (approved):
2 small NITs, but not critical.

docs/guides/anomaly-detection-tests/all-columns-anomalies.mdx
ellakz: these are lists of the test params, they're not all the same.
as for the whitespaces, maybe- I'll look into it after I merge :)

PR #1242, opened 2023-10-25
Backward compatibility for missing `raw_queries`

erikzaadi:
Ensure that `enriched_exposures` work even with older `dbt_exposures`

### DIFF
--- a/elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
+++ b/elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
@@ -6,14 +6,25 @@ config(
 )
 }}
 
-with dbt_exposures as (
-select * from {{ ref('dbt_exposures') }}
-),
-
-elementary_exposures as (
-select * from {{ ref('elementary_cli', 'elementary_exposures') }}
-)
-
-{# Union without duplicates where elementary_exposures has prio #}
-select * from dbt_exposures where dbt_exposures.unique_id not in (select elementary_exposures.unique_id from elementary_exposures)
-{{ elementary.sql_union_distinct() }} select * from elementary_exposures
+select
+    COALESCE(ee.unique_id, de.unique_id) as unique_id,
+    COALESCE(ee.name, de.name) as name,
+    COALESCE(ee.maturity, de.maturity) as maturity,
+    COALESCE(ee.type, de.type) as type,
+    COALESCE(ee.owner_email, de.owner_email) as owner_email,
+    COALESCE(ee.owner_name, de.owner_name) as owner_name,
+    COALESCE(ee.url, de.url) as url,
+    COALESCE(ee.depends_on_macros, de.depends_on_macros) as depends_on_macros,
+    COALESCE(ee.depends_on_nodes, de.depends_on_nodes) as depends_on_nodes,
+    COALESCE(ee.description, de.description) as description,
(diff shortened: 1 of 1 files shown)


elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
```diff
@@ -5,9 +5,23 @@ config(
...
 }}
+{% set dbt_exposures_relation = ref('elementary', 'dbt_exposures') %}
+{% set dbt_exposures_columns = adapter.get_columns_in_relation(dbt_exposures_relation) %}
+{% set raw_queries_column_exists = elementary.column_exists_in_relation(dbt_exposures_relation, 'raw_queries') %}
+{% set query %}
+    select
+    {% if raw_queries_column_exists %}
+        *
+    {% else %}
+        {% for column in dbt_exposures_columns %}
```
oravi, line 16: You know the schema here so I'd just expand it manually instead of doing the get_columns_in_relation which is probably slower as it runs describe on the warehouse

elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
```diff
@@ -5,9 +5,23 @@ config(
...
+        , NULL as raw_queries
+    {% endif %}
+{% endset %}
 
 with dbt_exposures as (
-select * from {{ ref('dbt_exposures') }}
+{{ query }} from {{ ref('dbt_exposures') }}
 ),
 
 elementary_exposures as (
```
oravi, line 27: Why not using join down here?

elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
```diff
@@ -5,9 +5,23 @@ config(
...
+        {% for column in dbt_exposures_columns %}
+            {{ column.name }},
+        {% endfor %}
+        , NULL as raw_queries
+    {% endif %}
+{% endset %}
 
 with dbt_exposures as (
-select * from {{ ref('dbt_exposures') }}
+{{ query }} from {{ ref('dbt_exposures') }}
```
oravi, line 24: I'd write the jinja of query right here inside the CTE

elementary/monitor/dbt_project/models/enriched_exposures/enriched_exposures.sql
```diff
@@ -5,15 +5,33 @@ config(
...
-
-elementary_exposures as (
-select * from {{ ref('elementary_cli', 'elementary_exposures') }}
-)
-
-{# Union without duplicates where elementary_exposures has prio #}
-select * from dbt_exposures where dbt_exposures.unique_id not in (select elementary_exposures.unique_id from elementary_exposures)
-{{ elementary.sql_union_distinct() }} select * from elementary_exposures
+select
+  {% for column in columns %}
```
oravi, line 32: Should it be columns_without_raw_queries instead of columns? also personally I'd prefer to put the list of columns explicitly here without the jinja for loop. I know it's not the most "engineering" way but in dbt they have this doc about best practices where they say that in SQL sometimes being clear (less jinja) is more important than DRY (do not repeat yourself) as the SQL becomes much clearer this way. Not critical, just worth mentioning this.
https://docs.getdbt.com/docs/build/jinja-macros#favor-readability-over--ness

oravi (approved):

PR #1243, opened 2023-10-25
groups: fix for exposures - path should be the path in the BI

haritamar:


elementary/monitor/api/groups/groups.py
```diff
@@ -53,6 +53,12 @@ def _update_dbt_group(
             return
 
         artifact_full_path_split = artifact.normalized_full_path.split(posixpath.sep)
+        if isinstance(artifact, NormalizedExposureSchema):
+            # For exposures, we want the path to be the path in the BI
+            artifact_full_path_split = artifact_full_path_split[
+                :2
+            ] + artifact.fqn.split("/")
```
haritamar, line 60: first two parts of the full path are the package name and "exposures", which we want to keep

NoyaArie: can we add a test?

NoyaArie (approved):
👍

elementary/monitor/api/groups/groups.py
haritamar: I'll actually replace with more straightforward logic

PR #1287, opened 2023-11-16
Allow specifying region_name for Boto3

jonboulle:
Add an `--aws-region-name` option to the CLI which gets passed through to the `boto3` client used by the S3 client.

When initialising a [Boto3 `Session`](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#session-configurations) it's useful to be able to specify the AWS Region to which the client should connect.

Context:

We're trying to use elementary to write to a [Cloudflare R2 bucket](https://developers.cloudflare.com/r2/), which [implements the S3 API](https://developers.cloudflare.com/r2/api/s3/api/). However, the S3 endpoints for these buckets do not encode the region, and hence the `--s3-endpoint-url` flag isn't sufficient to be able to configure elementary to successfully access the buckets.

Without this flag, it means we need write an AWS Profile configuration to disk and pass that in to elementary in order to override the region, which isn't desirable in contexts where we integrate elementary (e.g. invoking it from Airflow).


dapollak:
Hey @jonboulle, thanks for contributing to Elementary !
The change looks fine, except you must add the new cli option declaration so it will be possible to pass it as a cmd arg. See examples in `./elementary/monitor/cli.py`

jonboulle:
> Hey @jonboulle, thanks for contributing to Elementary ! The change looks fine, except you must add the new cli option declaration so it will be possible to pass it as a cmd arg. See examples in [cli.py](https://github.com/elementary-data/elementary/blob/master/elementary/monitor/cli.py)

Oh gosh, how did I miss that :facepalm: . Added!

dapollak (approved):

PR #1289, opened 2023-11-19
ELE-2020: Let error be None in PendingAlerts..

dapollak:


IDoneShaveIt (approved):
Thanks for the fix!
Don't forget to update SourceFreshnessAlertModel init to be optional as well!

PR #1292, opened 2023-11-22
ELE-2039 support multiple integrations at data monitoring alerts

IDoneShaveIt:


elementary/clients/slack/slack_message_builder.py
```diff
@@ -122,6 +122,20 @@ def create_button_action_block(text: str, url: str) -> dict:
             ],
         }
 
+    @staticmethod
+    def create_section_with_button(
```
ellakz, line 126: you don't need this one if you're not reformatting slack (this is intended to make the link be a button and not just hyperlink text)

elementary/monitor/data_monitoring/alerts/integrations/integrations.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self) -> None:
+        super().__init__(
+            "Failed to run edr monitor - no supported integration was provided"
+        )
+
+
+class Integrations:
+    @staticmethod
+    def create_integration(
```
ellakz, line 25: nitpicking 🤓 but maybe it's cause I'm  dealing a lot with ORM right now but I feel it should be "get" and not "create"

elementary/monitor/data_monitoring/alerts/integrations/slack/slack.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         self.tracking = tracking
         self.override_config_defaults = override_config_defaults
         self.message_builder = SlackAlertMessageBuilder()
-        super().__init__()
+        super().__init__(*args, **kwargs)
```
ellakz, line 68: do you sometimes pass on unknown arguments here? because if not I'd remove this entirely (from the init arguments as well)

elementary/clients/slack/slack_message_builder.py
IDoneShaveIt: Yeah but I don't mind adding this function as well because it is a generic one and could be used with the message builder everywhere 🙂

ellakz (approved):

PR #1340, opened 2023-12-26
ELE-2241 - add tags to report test results

IDoneShaveIt:


elementary/monitor/fetchers/tests/schema.py
```diff
@@ -25,9 +25,11 @@ class TestResultDBRowSchema(ExtendedBaseModel):
     test_results_description: Optional[str]
     owners: Optional[List[str]]
     model_owner: Optional[List[str]]
-    tags: Optional[List[str]]
+    tags: List[str] = Field(default_factory=list)
```
NoyaArie, line 28: do we still need this one?

elementary/monitor/api/tests/schema.py
```diff
@@ -62,6 +62,7 @@ class TestMetadataSchema(BaseModel):
     description: Optional[str] = None
     result: dict
     configuration: dict
+    test_tags: List[str] = Field(default_factory=list)
```
NoyaArie, line 65: maybe we should add this to `SourceFreshnessMetadataSchema` as well (just empty always). just because we are planning to use this field for all items in the table

NoyaArie (approved):
👍

elementary/monitor/api/tests/schema.py
IDoneShaveIt: Will do

elementary/monitor/fetchers/tests/schema.py
IDoneShaveIt: Well there is other place that uses it and this is the combination of model tags and test_tags.
I guess I will make it a property that return the union of the 2 🙂

PR #1341, opened 2023-12-26
Ele 2237 new filter schema

NoyaArie:


elementary/monitor/data_monitoring/schema.py
```diff
@@ -28,23 +28,48 @@ class ResourceType(Enum):
...
+
+class ResourceTypeFilterSchema(FilterSchema):
+    values: List[ResourceType]
+
+
+class FiltersSchema(BaseModel):
     selector: Optional[str] = None
     invocation_id: Optional[str] = None
-    invocation_time: Optional[str] = None
+    invocation_time: Optional[str] = Field(default=None)
```
IDoneShaveIt, line 56: Because the default field is not mutable so you can just use None 🙂 
The Field default factory is for dynamic defaults and mutable defaults like list and dict
```suggestion
    invocation_time: Optional[str] = None
```

IDoneShaveIt (approved):
Looking good!
Left one small pydantic comment 🙂

PR #1345, opened 2024-01-02
ELE-2263 - support multi filters at cli command

IDoneShaveIt:


elementary/monitor/cli.py
```diff
@@ -278,6 +288,7 @@ def monitor(
     suppression_interval,
     override_dbt_project_config,
     report_url,
+    filter,
```
ellakz, line 291: this is not very important but `filter` is a python builtin function and this means you are overriding it :\
can you set it to alerts_filter so it doesn't override? it can cause weird undebuggable bugs haha

elementary/monitor/cli.py
```diff
@@ -312,14 +323,25 @@ def monitor(
     anonymous_tracking.set_env("use_select", bool(select))
     try:
         config.validate_monitor()
-        selector_filter = SelectorFilter(config, anonymous_tracking, select)
+
+        filters = FiltersSchema()
+        if bool(filter):
```
ellakz, line 328: same here 🤓

elementary/monitor/data_monitoring/schema.py
```diff
@@ -101,6 +104,83 @@ def validate_report_selector(self):
...
+        if not cli_filters:
+            return FiltersSchema()
+
+        tags = []
+        owners = []
+        models = []
+        statuses = []
+        resource_types = []
+
+        for filter in cli_filters:
```
ellakz, line 118: same :)

elementary/monitor/data_monitoring/schema.py
```diff
@@ -101,6 +104,83 @@ def validate_report_selector(self):
...
+            tags_match = FiltersSchema._match_filter_regex(
+                filter_string=filter, regex=re.compile(r"tags:(.*)")
+            )
+            if tags_match:
+                tags.append(FilterSchema(values=tags_match))
+                continue
+
+            owners_match = FiltersSchema._match_filter_regex(
+                filter_string=filter, regex=re.compile(r"owners:(.*)")
```
ellakz, line 127: instead of using re.compile, re.search and groups, you can simply use `re.findall("models:(\w+)", filter)`. this returns a list, and if there's no match it's an empty list. will save you a lot of code

tests/unit/monitor/data_monitoring/test_filters_schema.py
```diff
@@ -0,0 +1,254 @@
...
+    assert len(filter_schema.statuses) == 1
+    assert sorted(filter_schema.statuses[0].values) == sorted(
+        ["fail", "error", "runtime error", "warn"]
+    )
+    assert len(filter_schema.resource_types) == 0
+
+    cli_filter = ("tags:tag1,tag2",)
+    filter_schema = FiltersSchema.from_cli_params(cli_filter)
+    assert len(filter_schema.tags) == 1
+    assert sorted(filter_schema.tags[0].values) == sorted(["tag1", "tag2"])
```
ellakz, line 35: so this means that it's multiple values (OR) and the test below is for AND?
Not sure if you did but it can be nice to add a comment explaining it in the FilterSchema class

ellakz (approved):

Issue #1353, opened 2024-01-05
`upload-source-freshness` can overflow CLI arguments max size, in some scenarios


willbowditch:
**Describe the bug**

On occasion the source freshness upload, `edr run-operation upload-source-freshness --project-dir . --profile-target dev` will exceed the maximum argument limit, with:
```
OSError: [Errno 7] Argument list too long: 'dbt'
```
In our deployment of Elementary this depends on the particular subset of sources we run, so I suspect the model name length and other metadata is occasionally going over the limit. 

**To Reproduce**

The nature of this error means it's will only be reproduced when the source arguments contains a lot of info. I can't share an example, but this could probably be simulated.

**Expected behavior**

It should chunk up the commands into sizeable batches.

**Environment (please complete the following information):**

- edr Version: 0.13.2
- dbt package Version: 0.13.0

**Additional context**

I believe the underlying cause is that Elementary currently chunks up the batches based on the number of records, into chunks of 100 records, regardless of record length. 

https://github.com/elementary-data/elementary/blob/3220d4b1f87ff52855b1f5d1ae1171cf652f603e/elementary/operations/upload_source_freshness.py#L61

I've got a patch which fixes this by calculating the character length and chunks using that. I will upload a PR shortly.

adamcunnington-mlg:
We are running into the same thing - except we have it every single time and it blocks us from using this functionality which is our main reason for choosing elementary!

haritamar:
Hi all, closing since now the default chunk size is lower and can be controlled with the `rows_per_insert` parameter.

The approach in the PR above is probably better but needs some adjustments (commented in more detail in the PR) - so feel free to re-open if you still feel it's needed.

PR #1354, opened 2024-01-05
Chunk calls to DBT based on string length

willbowditch:
Fixes issue https://github.com/elementary-data/elementary/issues/1353

## Bug

In some scenarios the argument list Elementary calls DBT with when uploading source freshness exceeds the limit of the operating system. 

This is because the chunking of calls to DBT is currently based on the number of sources and not the length of the arguments. So some combination of sources results in longer argument lists than others.

This PR changes the chunking to use the character length and adds a buffer to allow for a bit of variation. 

## Testing

I've tested this locally by patching elementary in our deployment and confirming the results are uploaded.

### DIFF
--- a/elementary/operations/upload_source_freshness.py
+++ b/elementary/operations/upload_source_freshness.py
@@ -12,8 +12,8 @@ from elementary.config.config import Config
 from elementary.monitor import dbt_project_utils
 from elementary.utils.ordered_yaml import OrderedYaml
 
-
 MAX_SERIALISED_CHARACTER_LENGTH = 90_000
+
 
 class UploadSourceFreshnessOperation:
     def __init__(self, config: Config):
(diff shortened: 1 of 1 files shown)


haritamar:
Hi @willbowditch !
Thank you for opening this PR and sorry for only just commenting on it.

I noticed that meanwhile we made a different change to upload_source_freshness that allows controlling the chunk size + lowering the default:
https://github.com/elementary-data/elementary/pull/1379

I know this is not the same, but the mechanisms are now conflicting.
If you feel this is still needed - I believe it makes sense to combine the two mechanisms so the `rows_per_insert` param would keep functioning (e.g. the default of it can be None, but if it's not None, then in any case don't create chunks with more rows than this value).

I'm closing this PR for now, but if you'd like please feel free to re-open with the changes above (or an alternative that makes sense)

PR #1356, opened 2024-01-05
move cloud and test dependencies to "extras"

cburroughs:
Most downstream users are not on multiple cloud providers, nor do they need to run the tests.  Allowing the option to select only what they need simplifies downstream dependency resolutions and prevents conflicts.

(I went with the dev- pattern tests instead of poetry groups, as the CONTRIBUTING instructions pointed towards using pip/requirements and not mandating poetry directly.)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 pytest
+pytest-parametrization>=2022.2.1
 pre-commit
 mypy
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,16 +21,15 @@ requests = ">=2.28.1,<3.0.0"
 beautifulsoup4 = "<5.0.0"
 ratelimit = "*"
 posthog = "<3.0.0"
-boto3 = "<2.0.0"
-google-cloud-storage = "<3.0.0"
+boto3 = {version = "<2.0.0", optional = true}
+google-cloud-storage = {version = "<3.0.0", optional = true}
 "ruamel.yaml" = "<1.0.0"
 alive-progress = "<=2.3.1"
 slack-sdk = ">=3.20.1,<4.0.0"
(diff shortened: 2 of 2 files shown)


cburroughs:
We use Python (surprise!) and support development on multiple CPU architectures (x86_64, arm64 and Operating Systems (Mac, Linux) and in turn support deployment to Linux on multiple CPU architectures.  To interact with the wider Python packaging ecosystem  we could:

  * (A) Say that everyone needs multiple native code compilers and library headers installed on their workstation, maintaining package lists for both Mac and a hodgepodge of Linux distros.
  * (B) Require all 3rd party python packages to be available as wheels, building the missing ones ourselves.
  
 (A) perhaps a more common choice but also an endless series of "IT" style support tarpits while (B) is closer to a problem amendable to software engineering, so we went with (B).  We are happy to spin up CI workers and build wheels as needed but in the case of things like test dependencies the idiomatic thing to do is usually  not to require them at all. (`pytest-parametrization` does not provide a wheel if that context was unclear.)
 
For the cloud provider case, due to the dynamic nature of python packaging pip uses a backtracking resolver with somewhat unpredictable performance characteristics.  Having less packages in the transitive graph thus always has marginally benefit and since most companies always use a single cloud provider, that's usually low hanging fruit for limiting scope.

Let me know that helps or if other context would be helpful.

haritamar:
Hi @cburroughs !
Sorry for the huge delay in our response, and thanks for your explanation above.

I think the points you made make sense. That being said, my primary concern is the same that @elongl brought above - which is breaking setup for existing Elementary users. 
We may do this separation eventually with an appropriate heads-up to users, but right now I feel the risk outweighs the benefit.

That being said, for test packages the decision is more straightforward, so I added only this change to a separate PR:
https://github.com/elementary-data/elementary/pull/1537

Closing this one.

PR #1370, opened 2024-01-14
Fix dbt deps logic in CLI

haritamar:


elementary/clients/dbt/dbt_runner.py
```diff
@@ -287,7 +290,22 @@ def _get_required_packages_names(self):
...
 
     def _run_deps_if_needed(self):
+        should_run_deps = False
+
         installed_package_names = set(self._get_installed_packages_names())
         required_package_names = set(self._get_required_packages_names())
         if not required_package_names.issubset(installed_package_names):
+            logger.info("Installing packages for edr internal dbt package...")
+            should_run_deps = True
+        else:
```
elongl, line 300: Make it an `elif`.

elementary/monitor/dbt_project_utils.py
```diff
@@ -9,42 +9,36 @@
...
 
-def _get_elementary_package_path():
-    package_path = os.path.join(_PACKAGES_PATH, _DBT_PACKAGE_NAME)
+def _get_elementary_package_path(project_dir: str) -> Optional[str]:
+    package_path = os.path.join(project_dir, "dbt_project", _DBT_PACKAGE_NAME)
     if os.path.exists(package_path):
         return package_path
 
-    legacy_package_path = os.path.join(_MODULES_PATH, _DBT_PACKAGE_NAME)
+    legacy_package_path = os.path.join(project_dir, "dbt_modules", _DBT_PACKAGE_NAME)
```
elongl, line 33: Let's deprecate it while we're at it.
(We already don't support _that_ old versions.)

haritamar: good point

elongl (approved):

Issue #739, opened 2023-03-06
Support dbt-trino connector


radium226:
**Is your feature request related to a problem? Please describe.**
At `$WORK`, I heavily leverage on [Trino](https://trino.io/) to use multiple data sources within my DBT projects (using the [`dbt-trino`](https://github.com/starburstdata/dbt-trino) connector). 

It's also quite useful for storing DBT's test results in a different database than the analytical one (by using the `--store-failures` flag combined with the `+schema` and `+database` config).

**Describe the solution you'd like**
Elementary is IMHO a better solution than the raw `--store-failures` flag and it could be great to be able to make it work with the `dbt-trino` connector. 

**Describe alternatives you've considered**
For now, none :) I'm still using the `--store-failures` flag but without the proper visualization, I can't make it very useful for the rest of the teams. 

**Additional context**
Nope.

**Would you be willing to contribute this feature?**
Totally! I started to dig a little bit (I seached `postgres__` to see what kind of macro needs to be implemented) but I need guidance on how to start and how to test.

Maayan-s:
Hi @radium226!
We would be happy to support your effort in making Elementary compatible with Trino. 
To be honest I'm not familiar with it so I don't know how hard it would be.

Generally speaking, we implemented every platform-specific functionality using the `adapter.dispatch` functionality, as dbt recommends. You can see an example in [this macro](https://github.com/elementary-data/dbt-data-reliability/blob/master/macros/utils/cross_db_utils/target_database.sql).
However, where there was a dbt_utils macro that we could use, we did. I do see utils in the dbt-trino adapter, so it looks promising in that sense.
Anyway, You can [see here](https://github.com/elementary-data/dbt-data-reliability/blob/master/macros/utils/cross_db_utils/safe_cast.sql) a workaround we did for a missing util in Spark.

I think you should approach it gradually -
Step 1 - Add support for uploading dbt artifacts and run results (in the dbt package).
Step 2 - Add support in the CLI for Slack alerts and UI generation.
Step 3 - Add support for data anomaly detection test (the most complex and platform-specific part of the code right now).

You can see here a guide for testing:
https://docs.elementary-data.com/general/contributions#contributing-to-the-dbt-package

leniartek:
Hello @radium226
Happy to hear you benefit from Trino! (dbt-trino PM here).

This is the first time I see a request for Elementary integration, I will take a closer look.

Thanks for reaching out.
Piotr

ndrluis:
[...]

PR #1378, opened 2024-01-17
ELE-489 Support dbt-trino connector

Tomme:
Relates to: https://github.com/elementary-data/elementary/issues/739

Other half: https://github.com/elementary-data/dbt-data-reliability/pull/652


elementary/monitor/dbt_project/macros/get_adapter_type_and_unique_id.sql
```diff
@@ -12,4 +12,8 @@
 
 {% macro bigquery__get_adapter_unique_id() %}
     {{ return(target.project) }}
-{% endmacro %}
\ No newline at end of file
+{% endmacro %}
+
+{% macro trino__get_adapter_unique_id() %}
+    {{ exceptions.raise_compiler_error("Force failure and thus set warehouse_type=null since Trino is not currently supported within the report index.html") }}
+{% endmacro %}
```
Tomme, line 19: Kinda strange but without this, `warehouse_type` will get populated within `elementary_output.json` and since `elementary/monitor/data_monitoring/report/index.html` doesn't know how to handle `"warehouse_type": "trino"` the report will fail to render, forcing a failure here will ensure `warehouse_type` is set to `null` which will allow the HTML to render.

This issue has not been seen with other adapters like Athena, since Athena will fallback to the `default__get_adapter_unique_id` macro and since it has no `host` property, it will also fail, resulting in `"warehouse_type": null`

Firstero:
Hi, I was wondering if the requirement related to this issue is still ongoing? Thank you!

Tomme:
> Hi, I was wondering if the requirement related to this issue is still ongoing? Thank you!

I'm unsure what you mean but from testing within my small to medium production environment everything is working as expected 😄

shezis:
Can we please expedite this much needed Component 🙏

elementary/monitor/dbt_project/macros/get_adapter_type_and_unique_id.sql
haritamar: I think we should probably fix this on the FE side on our end, I prefer this a bit to adding a hack, will look into it

haritamar (approved):

PR #1382, opened 2024-01-21
Alerts: Replace icon with color and make header informative

ellakz:


elementary/monitor/alerts/source_freshness_alert.py
```diff
@@ -153,8 +153,8 @@ def error_message(self) -> str:
     @property
     def summary(self) -> str:
         if self.original_status == "runtime error":
-            return f"Failed to calculate the source freshness of `{self.source_name}`"
-        return f"Freshness exceeded the acceptable times on source `{self.source_name}`"
+            return f'Failed to calculate the source freshness of "{self.source_name}"'
```
IDoneShaveIt, line 156: Do we want to change it from `this format` to "just quotes"?

elementary/monitor/alerts/source_freshness_alert.py
```diff
@@ -153,8 +153,8 @@ def error_message(self) -> str:
...
+        return f'Freshness exceeded the acceptable times on source "{self.source_name}"'
```
IDoneShaveIt, line 157: Same 🙂

elementary/monitor/data_monitoring/alerts/integrations/slack/message_builder.py
```diff
@@ -78,6 +78,12 @@ def add_details_to_slack_alert(
             ]
             self._add_blocks_as_attachments(configuration_blocks)
 
+    def add_color_to_slack_alert(self, color: str):
```
IDoneShaveIt, line 81: Maybe we should move it to the generic `SlackMessageBuilder`?
It feels like adding color is a generic format that could be used in other places as well 🙂

IDoneShaveIt: And we will need to rename the method 🙂

IDoneShaveIt: Maybe add a comment for the color string format?
Just to make it easier for new ones 🙂

elementary/monitor/data_monitoring/alerts/integrations/slack/message_builder.py
```diff
@@ -78,6 +78,12 @@ def add_details_to_slack_alert(
...
+        for block in self.slack_message.get("blocks", []):
```
IDoneShaveIt, line 82: I know that is not part of this change, but this is small and could be really nice for the future, so if you feel like doing so it would be great!

Maybe we can add a @property to the `SlackMessageBuilder` for `blocks` and `attachments` that returns `slack_message.get("blocks", [])/slack_message.get("attachments", [])`

tests/unit/clients/slack/test_slack_message_builder.py
```diff
@@ -215,15 +215,6 @@ def test_create_compacted_sections_blocks():
     )
 
 
-def test_get_slack_status_icon():
```
IDoneShaveIt, line 218: We removed this test, but maybe we can add a new one that checks that we add the color properly?

IDoneShaveIt (commented):
Looking great!
Added small comments 🙂

elementary/monitor/alerts/source_freshness_alert.py
ellakz: Since the only place it's currently used is the title, and the title is plain test, this had to be done

IDoneShaveIt (approved):
LGTM!

PR #1422, opened 2024-02-20
Handle sent_at column at alerts_v2 contains timezone

IDoneShaveIt:


elementary/monitor/data_monitoring/alerts/data_monitoring_alerts.py
```diff
@@ -128,7 +129,10 @@ def _get_suppressed_alerts(
             )
             last_sent_time = alerts_last_sent_times.get(alert_class_id)
             is_alert_in_suppression = (
-                (current_time_utc - last_sent_time).total_seconds() / 3600
+                (
+                    current_time_utc - convert_time_to_timezone(last_sent_time)
```
ellakz, line 133: what happens if last_sent_time is on another timezone and not UTC? won't that cause bugs?

IDoneShaveIt: So if the db adds timezone to the timestamp column it will have +/- time.
So the `convert_time_to_timezone` will parse the timezone and convert it to the relevant utc time and it will work normally 🙂

ellakz: Approving, and assuming you checked that :)

ellakz (approved):

PR #1426, opened 2024-02-22
support pydantic 2 in elementary

ellakz:


elementary/utils/pydantic_shim.py
```diff
@@ -0,0 +1,32 @@
+"""
```
ellakz, line 1: this is the main interesting change

mypy.ini
```diff
@@ -14,3 +14,9 @@ ignore_missing_imports = True
 
 [mypy-posthog]
 ignore_missing_imports = True
+
+[mypy-pydantic]
```
ellakz, line 18: not sure if there was another way to overcome mypy issues (it was angry at whichever version wasn't installed for my when I ran it)

elementary/monitor/api/models/schema.py
```diff
@@ -38,17 +37,17 @@ def format_normalized_full_path_sep(cls, normalized_full_path: str) -> str:
 
 # NormalizedArtifactSchema must be first in the inheritance order
 class NormalizedModelSchema(NormalizedArtifactSchema, ModelSchema):
-    artifact_type: str = Field("model", const=True)
+    artifact_type: str = Field("model", const=True)  # type: ignore  # noqa
```
ellakz, line 40: unfortunately Field in pydantic 2 doesn't have a "const" field and that causes mypy to error. ignoring until we migrate to pydantic 2 for real

haritamar (approved):

PR #1431, opened 2024-02-25
ELE-2569 - fix source freshness alerts created_at and updated_at are None

IDoneShaveIt:


tests/mocks/fetchers/alerts_fetcher_mock.py
IDoneShaveIt, line 1: Moving code to make it importable from tests - no new lines

elementary/monitor/fetchers/alerts/schema/pending_alerts.py
```diff
@@ -51,6 +51,19 @@ class Config:
         # Make sure that serializing Enum return values
         use_enum_values = True
 
+    @root_validator(pre=True)
+    def validate_times(cls, values: dict) -> dict:
+        new_values = {**values}
```
dapollak, line 56: why not just set `datetime.utcnow()` as default for those fields ?

IDoneShaveIt: `updated_at: datetime = Field(default_factory=datetime.utcnow)` Will fail on validation when updated_at is None (which this is the case we want to solve).

Setting `updated_at: Optional[datetime] = Field(default_factory=datetime.utcnow)` will keep updated_at as None.

So if we want to be accept updated_at as None and parse it to utcnow, we need to use a validator 🙂

haritamar (approved):

PR #1443, opened 2024-03-04
ELE-2617 - Fix disable samples for alerts

IDoneShaveIt:
Make sure we disable the alerts samples when the flag --disable-samples is True

### DIFF
--- a/elementary/monitor/fetchers/alerts/schema/alert_data.py
+++ b/elementary/monitor/fetchers/alerts/schema/alert_data.py
@@ -105,6 +105,8 @@ class BaseAlertDataSchema(BaseModel):
         elementary_database_and_schema: Optional[str] = None,
         global_suppression_interval: int = 0,
         override_config: bool = False,
+        *args,
+        **kwargs
     ):
         raise NotImplementedError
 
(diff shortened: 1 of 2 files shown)


erikzaadi (approved):

elementary/monitor/fetchers/alerts/schema/alert_data.py
```diff
@@ -183,6 +185,9 @@ def format_alert(
         elementary_database_and_schema: Optional[str] = None,
         global_suppression_interval: int = 0,
         override_config: bool = False,
+        disable_samples: bool = False,
+        *args,
```
ellakz, line 189: why do you need those? it's better to be explicit about what args and kwargs can be passed on

IDoneShaveIt: The function that is using the `format_alert` method is calling it for 3 different type of alerts.
Each alert has different arguments for the `format_alert` method.
That way we can assure that each `format_alert` define only the arguments it needed, and not fail when we pass it extra arguments.

This is (as I see it) a valid use of args and kwargs 🙂

PR #1451, opened 2024-03-06
fix: Fixed update_sent_alerts macro

svdimchenko:
Problem statement:
When running with dbt-athena adapter I get the error on Athena side:

```
TYPE_MISMATCH: line 3:17: UPDATE table column types don't match SET expressions: Table: [varchar, timestamp(6), timestamp(6)], Expressions: [varchar(4), varchar(19), timestamp(3) with time zone]
```

and the query is following:

```
-- /* {"app": "dbt", "dbt_version": "1.7.9", "profile_name": "elementary", "target_name": "clickstream", "node_id": "model.elementary_cli.update_sent_alerts"} */

                UPDATE "awsdatacatalog"."elementary_clickstream"."alerts_v2" set status = 'sent', sent_at = '2024-03-06 14:25:04', updated_at = CURRENT_TIMESTAMP
                WHERE alert_id IN 
        ( 'e910deb7-46c8-4de1-a73f-c5143d6fa5d5.model.clickstream.growthbook_casino_results'  )
 and status = 'pending' and
                    coalesce(
        try_cast(detected_at as timestamp),
        cast(from_iso8601_timestamp(cast(detected_at AS varchar)) AS timestamp)
    ) >= coalesce(
        try_cast(date_add('day', -31, date_trunc('day', CURRENT_TIMESTAMP)) as timestamp),
        cast(from_iso8601_timestamp(cast(date_add('day', -31, date_trunc('day', CURRENT_TIMESTAMP)) AS varchar)) AS timestamp)
    )
```

Current PR fixes the issue.


elementary/monitor/dbt_project/macros/alerts/manage/update_sent_alerts.sql
```diff
@@ -2,11 +2,15 @@
     {% if execute %}
         {% if alert_ids %}
             {% set update_sent_alerts_query %}
-                UPDATE {{ ref('elementary_cli', 'alerts_v2') }} set status = 'sent', sent_at = {{ elementary.edr_quote(sent_at) }}, updated_at = {{ elementary.edr_current_timestamp() }}
-                WHERE alert_id IN {{ elementary.strings_list_to_tuple(alert_ids) }} and status = 'pending' and
-                    {{ elementary.edr_cast_as_timestamp('detected_at') }} >= {{ elementary_cli.get_alerts_time_limit() }}
+                update {{ ref('elementary_cli', 'alerts_v2') }}
+                set status = 'sent',
+                    sent_at = {{ elementary.edr_cast_as_timestamp(elementary.edr_quote(sent_at)) }},
```
IDoneShaveIt, line 7: Hey, thank you for raising that up!
The reason `sent_at` was set to `edr_quote` and not a timestamp is because the old tables that we were using were setting it to `edr_quote` and not timestamp.

I have no problem with this change as it makes sense, but we need to make sure that we don't break things here for other adapters / backwards compatibility / users that already has sent_at as `edr_quote`

IDoneShaveIt (changes requested):
Thank you for the contribution and raising that error!
We need to think a bit about how it will affect other adapters and users that their table is already using sent_at as string and not timestamp 🙂

elementary/monitor/dbt_project/macros/alerts/manage/update_sent_alerts.sql
IDoneShaveIt: We had a similar situation in the [following PR](https://github.com/elementary-data/elementary/pull/1318/files).
You can use the same implementation here - if relation already exists, check the type of the `sent_at` column and use it (could be string or timestamp). If not, use timestamp 🙂

svdimchenko: I'll take a look there. However as I see from ddl, field sent_at is of `timestamp` data type:
```
{% macro empty_alerts_v2() %}
    {{ elementary.empty_table([
		('alert_id', 'long_string'),
		('alert_class_id', 'long_string'),
		('type', 'string'),
		('detected_at','timestamp'),
    	('created_at','timestamp'),
    	('updated_at','timestamp'),
		('status','string'),
		('data', 'long_string'),
		('sent_at', 'timestamp')
    ]) }}
{% endmacro %}
```

IDoneShaveIt: Missed that!
So it might be all good to implement that change without the backwards support.
Let me check that and get back to you 🙂

svdimchenko: Hi there @IDoneShaveIt ! Just kind reminder: can we move further with this PR ?

IDoneShaveIt: Hey @svdimchenko, sorry for the late response (I was away).
I checked it and it seems like your solution will work (as we already using the column as timestamp as you mentioned) 🙂 

Approving the PR, thank you again for your contribution!

IDoneShaveIt (approved):

PR #1485, opened 2024-04-11
remove test subtype from most alerts

ellakz:


elementary/monitor/alerts/test_alert.py
```diff
@@ -152,22 +152,28 @@ def data(self) -> Dict:
 
     @property
     def concise_name(self) -> str:
-        if self.is_elementary_test:
+        if self.test_sub_type_display_name.lower() not in (
```
IDoneShaveIt, line 155: `concise_name` is used on the group_by method both in Slack and Teams.
Make sure that this change is not breaking anything 🙏🏻

elementary/monitor/alerts/test_alert.py
```diff
@@ -152,22 +152,28 @@ def data(self) -> Dict:
...
             return f"{self.test_short_name or self.test_name}"
 
     @property
-    def summary(self) -> str:
-        if self.test_type == "schema_change":
-            return (
-                f"{self.test_sub_type_display_name} on "
-                f"{self.table_full_name + '.' + self.column_name if self.column_name else self.table_full_name}"
-            )
+    def asset_name(self) -> str:
```
IDoneShaveIt, line 165: `asset_name` should not be a property as it is not used outside of the `summary` property.
Instead it should be a var you define inside the `summary` property and use it there (as the only reason you created this property, if I understand it correctly, is to not reuse code inside `summary` 🙂

elementary/monitor/alerts/test_alert.py
ellakz: yeah I want to remove this unnecessary thing in the group alerts too

IDoneShaveIt (approved):
lgtm

Issue #1421, opened 2024-02-19
Incompatibility elementary-data with Airflow (MWAA)


mattxxi:
**Describe the bug**

We are currently using Airflow (MWAA) 2.7.2, we would like to use elementary with Airflow and run edr commands in operators.
But there is a constraint for the library pydantic which has to be 2.4.2 but elementary-data is using pydantic <2.0.0.

*Please find all airflow constraints here:*
https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.11.txt

**To Reproduce**

Create a file requirements.txt with:
--constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.11.txt"
elementary-data==0.14.0

Then run the command:
```shell
pip download -d tmp/plugins --requirement requirements.txt
```
You must obtain the following error:
![image](https://github.com/elementary-data/elementary/assets/36408416/b94771d7-d1be-480b-b13c-bb9fc3c201a1)

**Expected behavior**

We should be able to install all requirements without conflicts between airflow and elementary.

**Environment (please complete the following information):**

- edr Version: 0.14.0
- dbt package Version: 0.14.0
- airflow: 2.7.2

joydeepml:
One way to get around conflicting requirements between Airflow and dbt/packages is to run dbt inside a virtualenv within the airflow workers. Can set up a virtualenv using a startup script and activate it before running the dbt commands.

kokorin:
> One way to get around conflicting requirements between Airflow and dbt/packages is to run dbt inside a virtualenv within the airflow workers. Can set up a virtualenv using a startup script and activate it before running the dbt commands.

That's exactly what we do.

haritamar:
Hi @mattxxi !
Thanks for opening this issue and apologies for only replying now.
The issue with the pydantic requirement has actually been solved since version 0.14.1 of the package!

(Running in a virtualenv though as suggested above is a good practice in general, if you are already doing so)

PR #1489, opened 2024-04-12
Loosen the version constrain of the click package

yu-iskw:
## Motivation
I got the error message to install elementary in addition to dbt-core. I assume that is caused by the difference of the version constrains of the click package in dbt-core and elememtary. I would expect pip appropriately handle the version constrains in both of dbt-core and elementary. But, sometimes it doesn't work. In my opinion, we are probably loosen the version constrains so that we can also catch up with new releases of click. The latest version is today 8.1.7.

- pip: 23.3.2
- dbt-core: 1.7.11

```shell
$ pip install 'elementary-data[bigquery]'
...
elementary-data 0.14.0 requires click<8.1.4,>=7.0, but you'll have click 8.1.7 which is incompatible.
```

https://github.com/dbt-labs/dbt-core/blob/c86cec3256cfeb636d47f21279f3cb689d585439/core/setup.py#L62


yu-iskw:
@oravi Sorry for bothering your. What are your thoughts on the change?

mi-volodin:
Upvoting this one. 
Elementary-data now in direct conflict with Airflow 2.7.2 [constraints](https://github.com/apache/airflow/blob/1416d3662da4e61e5b9b72accc5bd3c21e042368/constraints-3.8.txt#L256) where click has == 8.1.7 version.

List of changes [for the click](https://click.palletsprojects.com/en/8.1.x/changes/#version-8-1-7) since 8.1.4, seems only issue fixes and no breaking changes.

mi-volodin:
@haritamar Can you take a look at this? I saw you closed a[ pydantic issue,](https://github.com/elementary-data/elementary/issues/1421) which is similar (I wonder how it wasn't mentioned in the same issue).

haritamar (approved):

haritamar:
Hi @mi-volodin  and @yu-iskw !
Sorry for the delay here.
The limitation was meant to be temporary due to an issue in v8.1.4 specifically - but we can totally loosen it now.

mi-volodin:
@haritamar thanks a lot! A great relief for the team.
@yu-iskw thanks for making a PR 🙏

yu-iskw:
@haritamar Thank you for merging this.

yu-iskw:
@haritamar Can we release a new version which includes the changes?

PR #1490, opened 2024-04-14
Refactor Slack integration and Slack Alert message builder to make them editable easily

IDoneShaveIt:
- Templates now return pydantic schema
- Converting template to Slack message is now happening as part of `_get_alert_template`


noaKurman (approved):
LGTM

PR #1501, opened 2024-04-25
Filter exposures with no upstream tables in lineage.

elongl:


elementary/monitor/api/models/models.py
```diff
@@ -175,6 +186,23 @@ def get_test_coverages(self) -> Dict[str, ModelCoverageSchema]:
                 )
         return coverages
 
+    def _exposure_has_upstream_node(
```
elongl, line 189: Shall I add caching to this function?

ofek1weiss (approved):

PR #1519, opened 2024-05-08
remove refs to information_schema_columns

haritamar:


elementary/monitor/dbt_project/models/tests_recommendation/table_timestamp_columns.sql
```diff
@@ -41,7 +41,7 @@ with
             lower(schema_name) as schema_name,
             lower(table_name) as table_name,
             lower(column_name) as column_name
-        from {{ ref("elementary", "information_schema_columns") }}
+        from {{ elementary.get_elementary_relation('information_schema_columns') }}
```
haritamar, line 44: For now just replaced the ref with `get_elementary_relation` so compilation won't fail

IDoneShaveIt (approved):

PR #1520, opened 2024-05-09
ELE-3044: New dimensions features docs

dapollak:


elongl (approved):

docs/data-tests/anomaly-detection-configuration/exclude-final-results.mdx
```diff
@@ -0,0 +1,34 @@
...
+sidebarTitle: "exclude_final_results"
+---
+
+`exclude_final_results: [SQL where expression on fields value / average]`
+
+Failures in dimension anomaly tests consist of outliers in row counts across all dimensions during the training period.
+Some dimensions may contribute metrics that are considered insignificant compared to others, and you may prefer not to receive alerts for them.
+With this parameter, you can disregard such failures.
+
+1. `value` - Outlier row count of a dimension during the detection period.
```
ellakz, line 12: did we add this only to the dimension anomalies or to the column anomalies as well? because "value" in column anomalies isn't row count (and also average isn't), which actually leads me to believe we need these metrics to always be the row count even when the metric isn't (in metrics like string length the value doesn't imply the significance)

PR #1549, opened 2024-06-06
add optional --hours-back for edr monitor

miktros:
Add optional --hours-back for edr monitor to specify how many hours back should edr monitor look for pending alerts.

### DIFF
--- a/elementary/monitor/api/alerts/alerts.py
+++ b/elementary/monitor/api/alerts/alerts.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from elementary.clients.api.api_client import APIClient
 from elementary.clients.dbt.dbt_runner import DbtRunner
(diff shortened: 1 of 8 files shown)


ellakz:
Hi @miktros 
I'm not sure how this PR solves the issue in question.
Commented on the issue as well but I believe this will not solve your problem.
If you could shed some light on it I would be happy to review.
Thanks!

miktros:
By splitting data into 1-hour time buckets in volume.anomalies tests, we were hoping to be able to run `dbt test` and then `edr monitor` every hour so that alert is emitted only for a test with row count in the most recent 1-hour time bucket that is lower than past 21 day average (time bucket failure). We use the "hours-back" option, `edr monitor --hours-back 1`, to achieve the alerting behavior as described above. With the "days-back" option, alert is emitted each time `edr monitor` is run when there is a time bucket failure any time during the day, even though the most recent time bucket has no failure. In short, we want per hour alert notification that reflects failure condition of the most recent hour.

mstfkmlbsbdk2023:
if I don't add any parameter as --hours-back. it's getting 1 as default. is that correct or is there any configuration to change it?

here is the command how to run edr monitor: edr monitor --teams-webhook<> --group-by table

Screenshot 2024-07-09 at 09 24 04

PR #1673, opened 2024-08-15
Populate seed alterts together with models and snapshots

### LINKED ISSUE #1654: Support seed monitoring with EDR
In our project we occasionally got an error with DBT seed (new column was added). 
We use the following EDR command:

```sh
edr monitor --profile-target ${DBT_TARGET} --env ${DBT_TARGET} --target-path edr_target \
                  --filters statuses:error,fail --filters resource_types:model,test --teams-webhook ${TEAMS_ALERT_WEBHOOK}
```

Unfortunately EDR fails with `ValueError: 'seed' is not a valid ResourceType` if we add `seed` to resource_type filter.

EDR allows sending seed failure/error notifications.

**Describe alternatives you've considered**
Nothing

**Additional context**
I tried adding `SEED` value to `ResourceType` enum, the command completed successfully but no notification was sent. 
`alerts_v2` model's post-hook checks only for `test_alerts`, `model_alerts` and `source_freshness_alerts`. There is no seed alerts.

**Would you be willing to contribute this feature?**
Yes with guidance.

kokorin:
Closes #1654


elementary/monitor/dbt_project/macros/alerts/population/model_alerts.sql
```diff
@@ -61,63 +67,39 @@
...
-            original_path,
-            owner,
-            alias 
-        from model_run_results
-        where lower(status) != 'success'
-        and {{ elementary.edr_cast_as_timestamp('generated_at') }} > {{ elementary.edr_timeadd('day', -1 * days_back, elementary.edr_current_timestamp()) }}
-        union all
-        select 
-            model_execution_id,
+    seed_run_results as (
```
elongl, line 70: Can you make sure to render those only in the `seed_run_results` actually exists?
You can see an example of how to do so [here](https://github.com/elementary-data/elementary/blob/7d3deefb5e690822a2e791c7310aaf5072983ffb/elementary/monitor/dbt_project/models/alerts/deprecated/alerts_models.sql#L13).

Essentially define the relation at the top of the model, and then wrap its usages with `{% if seed_run_results_relation %}`.

haritamar: We shouldn't add it, as the safety is intended to handle cases where an older dbt package version is used (that doesn't yet have `seed_run_results`)

haritamar (approved):
looks good I think, merging

PR #1674, opened 2024-08-19
allow disabling elementary version check

slaupster:
checking the version is nice and all, but when running in production it just creates firewall log spam - we don't let traffic out anywhere, everything is explicitly allowed in kubernetes overlay network policy as well as an overarching firewall policy. Everytime elementary runs we see the spam trying to contact pypi to check for a newer version.

This PR creates a simple config to allow disabling the check.


elementary/cli/upgrade.py
```diff
@@ -3,23 +3,27 @@
...
 def recommend_version_upgrade():
-    try:
-        latest_version = package.get_latest_package_version()
-        current_version = package.get_package_version()
+    config = Config()
 
-        if not latest_version:
-            # Failed to obtain the latest version, so skip the check
-            return
+    if not config.disable_elementary_version_check:
```
elongl, line 13: Maybe change this so that there wouldn't be a need to indent the rest of the code:

```py
    if config.disable_elementary_version_check:
        return
```

elongl:
Thanks a lot for this contribution, makes a lot of sense!
Left one small comment, will merge afterwards.

elementary/cli/upgrade.py
slaupster: thanks @elongl. Sorry for late reply, I've been on holiday.

This is done.

PR #1697, opened 2024-09-08
Test models in the report

MikaKerman:
A new "tests" section in the JSON report, with pure test data, was added to make the test metadata (now called "test") accessible even when there were no runs and results.


elementary/monitor/dbt_project/macros/base_queries/tests.sql
```diff
@@ -1,56 +0,0 @@
-{% macro get_tests() %}
```
ofek1weiss, line 1: this was the unused macro right?

ofek1weiss (approved):

elementary/monitor/dbt_project/macros/base_queries/tests.sql
MikaKerman: correct, but please double check me 🙏

PR #1747, opened 2024-11-13
fix indentation in dbt tests alerts code examples

japerry911:
Hello team,

I was receiving the follow error with current indentation examples:

```
Parsing Error
  Invalid test config given in models/staging/airbyte_salesforce/stg__airbyte_salesforce_account.yml:
        test definition dictionary must have exactly one key, got [('not_null', None), ('meta', {'description': 'Not NULL test for Fax.'})] instead (2 keys)
        @: UnparsedModelUpdate(original_file_path='mode...ne)
```

I have updated the applicable spots.

Also, when should `tests` be updated to `data_tests` in docs? I know we use `data_tests` now with  1.8.x dbt core.

Thank you 🚀

### DIFF
--- a/docs/_snippets/guides/alerts-code-configuration.mdx
+++ b/docs/_snippets/guides/alerts-code-configuration.mdx
@@ -62,8 +62,8 @@ models:
 ```yml test
 tests:
   - not_null:
-    meta:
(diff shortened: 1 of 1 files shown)


haritamar (approved):

haritamar:
Thanks for the fix @japerry911 !

PR #1765, opened 2024-12-18
added snapshots to the report

MikaKerman:
null


MikaKerman:
Duplication of another merged PR

PR #1795, opened 2025-02-03
Ele 3960 adaptive cards

MikaKerman:
null

### DIFF
--- a/elementary/messages/blocks.py
+++ b/elementary/messages/blocks.py
@@ -109,4 +109,5 @@ InExpandableBlock = Union[
     "ExpandableBlock",
 ]
 
-ExpandableBlock.model_rebuild()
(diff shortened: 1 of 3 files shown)


elementary/messages/blocks.py
```diff
@@ -24,7 +25,22 @@ class TextStyle(Enum):
     ITALIC = "italic"
 
 
+class BlockType(Enum):
```
ofek1weiss, line 28: i think i would just do `str`
its used only for pydantic, so there wont be an issue of comparing to raw text, and it will probably become a kader if and when we want to add cloud specific blocks

MikaKerman: Done.

ofek1weiss (approved):

PR #1796, opened 2025-02-03
Ele 4028 format alerts messages

MikaKerman:
null


MikaKerman:
I have reduced the number of test cases.

MikaKerman:
@ofek1weiss 
Also added an env to the alert, in case it was specified with "--env X" in the cli

PR #1798, opened 2025-02-04
Ele 4029 slack message blocks

ofek1weiss:


tests/unit/messages/formats/adaptive_cards/test_adaptive_cards.py
```diff
@@ -10,28 +10,13 @@
```
ofek1weiss, line 10: moved all test logic to `base_test_format.py`

tests/unit/messages/formats/base_test_format.py
```diff
@@ -0,0 +1,327 @@
+from abc import abstractmethod
```
ofek1weiss, line 1: adapted from `test_adaptive_cards.py`

elementary/messages/formats/block_kit.py
```diff
@@ -0,0 +1,231 @@
...
+                secondary_facts = []
+                self._add_primary_fact(current_fact)
+            else:
+                if len(secondary_facts) >= self._SECONDARY_FACT_CHUNK_SIZE:
+                    self._add_secondary_facts(secondary_facts)
+                    secondary_facts = []
+                secondary_facts.append(current_fact)
+        self._add_secondary_facts(secondary_facts)
+
+    def _add_divider_block(self, block: DividerBlock) -> None:
```
MikaKerman, line 156: Is it necessary to get the block?

MikaKerman (approved):

MikaKerman:
@ofek1weiss What about adding tests to the alerts message builder?

elementary/messages/formats/block_kit.py
ofek1weiss: its not necessary, but it think it makes more sense for this function to decide that it isn't

ofek1weiss:
> @ofek1weiss What about adding tests to the alerts message builder?

The message builder did not exist at the time, will do

Issue #1808, opened 2025-02-10
Group alerts by owner


szaffarano:
**Is your feature request related to a problem? Please describe.**

Our DWH has many teams developing models, and it would be nice to group alerts by owners so that each team is tagged once and has a summary of the alerts. Then, they can click on the report link to have a more detailed overview. Grouping by model is not enough since they are tagged many times, so they have to go through each message individually.

**Describe the solution you'd like**

The solution would be adding a new grouping option to the `edr monitor` command, `--group-by owner`, e.g.:

```bash
edr monitor \
  --slack-token ... \
  --slack-channel-name ... \
  --profile-target ... \
  --report-url ...
  --group-by owner \
  --test true
```

**Describe alternatives you've considered**

AFAIK, the current implementation does not allow what we want. Another option would be to add a new `edr` parameter allowing some sort of message templating so that the messages can be built customarily, but this also needs to be added to `edr`, and I think it's a more complex solution.

**Additional context**

Proposed message:

![Image](https://github.com/user-attachments/assets/6793a337-8b0d-46e3-ad61-60325c288b83)

Where the `View tests runs` link opens the report selecting the tests filtering by owner. And `See more` shows the detail, grouped by model:

![Image](https://github.com/user-attachments/assets/99ed5eaf-862c-4ea8-b822-49f1320ea3f3)

**Would you be willing to contribute this feature?**

Absolutely!

NoyaOffer:
We're checking out the PR!

PR #1809, opened 2025-02-10
feat: Group alerts by owner

szaffarano:
This is an initial implementation of what I proposed in https://github.com/elementary-data/elementary/issues/1808. It is only focused on `Slack`, but if you agree with what I did, I could also implement the `Teams` support.

- [x] Slack implementation
- [ ] Teams implementation (still WIP)
- [ ] Add tests

### DIFF
--- a/elementary/messages/formats/adaptive_cards.py
+++ b/elementary/messages/formats/adaptive_cards.py
@@ -16,22 +16,8 @@ from elementary.messages.blocks import (
     TextBlock,
     TextStyle,
 )
+from elementary.messages.formats.html import ICON_TO_HTML
 from elementary.messages.message_body import Color, MessageBlock, MessageBody
-
-ICON_TO_HTML = {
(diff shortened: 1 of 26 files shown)


elementary/config/config.py
```diff
@@ -88,11 +88,11 @@ def __init__(
 
         config = self._load_configuration()
 
-        self.target_dir = self._first_not_none(
+        self.target_dir = str(self._first_not_none(
```
szaffarano, line 91: the changes here were because of some linting warnings from pyright I got

elementary/monitor/data_monitoring/alerts/integrations/teams/teams.py
szaffarano, line 1: Here is a WIP because I don't have a valid `Teams` user to test it, but I can finish it if you agree with the current implementation.

PR #1828, opened 2025-02-18
Ele 4074 selectable alert sections

ofek1weiss:
null

### DIFF
--- a/elementary/monitor/alerts/alert_messages/builder.py
+++ b/elementary/monitor/alerts/alert_messages/builder.py
@@ -28,6 +28,7 @@ from elementary.messages.blocks import (
     TextStyle,
 )
 from elementary.messages.message_body import Color, MessageBlock, MessageBody
+from elementary.monitor.alerts.alert_messages.alert_fields import AlertFields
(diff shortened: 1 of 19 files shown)


elementary/monitor/alerts/alert_messages/builder.py
```diff
@@ -274,7 +284,9 @@ def _get_result_blocks(
             result_blocks.append(
                 CodeBlock(text=result_message.strip()),
             )
-        if result_sample:
+        if (
```
ofek1weiss, line 287: unrelated - made more aligned with the way it was before: anomalous value should be considered as the result sample if present

elementary/monitor/alerts/alert_messages/builder.py
```diff
@@ -203,13 +205,13 @@ def _get_details_blocks(
             )
         )
         fact_blocks = []
-        if table:
+        if table and AlertField.TABLE in fields:
```
MikaKerman, line 208: Wouldn't it make more sense to move all these conditions to the alerts side?

MikaKerman (approved):

PR #1701, opened 2024-09-10
GH-1699 Setup multi-platform Docker images to support both x86 and arm architectures

JanDintel:
**Introduction**

Support multiple platforms (architectures) for the Docker container in the Github Action release.

In this way open-source users can run the Docker image on both `x86` and `arm` architectures. Rational behind this is because more cloud providers are supporting `arm` architectures with provide a better price-performance.

Closed GH-1699

**Supporting documentation**

- https://github.com/marketplace/actions/build-and-push-docker-images
- https://docs.docker.com/build/ci/github-actions/multi-platform/

PR #1830, opened 2025-02-18
Revert multi-arch build support to avoid QEMU segfaults

MikaKerman:
Reverts elementary-data/elementary#1701

This PR reverts the changes that added support for the `linux/arm64` architecture in our Docker builds. We’ve been encountering recurring segmentation faults under QEMU when installing certain packages (specifically `libsasl2-dev`), causing the arm64 build to fail. 

Since we do not strictly need an arm64 image at this time, this revert removes the multi-arch workflow steps and returns us to a straightforward `amd64`-only build. By reverting these changes, we avoid the QEMU user-mode emulation crashes and restore reliable builds. 

We can reconsider arm64 support in the future if it becomes necessary or if we move to a native arm64 builder.


haritamar (approved):

MikaKerman:
Hey @JanDintel,
We had to revert the [PR](https://github.com/elementary-data/elementary/pull/1701) due to QEMU segfaults on arm64 that occur whenever certain packages trigger a libc reconfiguration.
We added a new image build workflow, allowing testing of the build before the package release process.

PR #1846, opened 2025-03-09
Consider invocation filters in the send-report summary

ofek1weiss:
null

### DIFF
--- a/elementary/monitor/api/tests/tests.py
+++ b/elementary/monitor/api/tests/tests.py
@@ -73,12 +73,6 @@ class TestsAPI(APIClient):
         dbt_invocation: Optional[DbtInvocationSchema] = None,
     ) -> List[TestResultSummarySchema]:
         filtered_test_results_db_rows = self.test_results_db_rows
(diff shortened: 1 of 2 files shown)


elementary/monitor/api/tests/tests.py
```diff
@@ -69,8 +70,15 @@ def _get_test_results_db_rows(
...
         filter: SelectorFilterSchema = SelectorFilterSchema(),
+        dbt_invocation: Optional[DbtInvocationSchema] = None,
     ) -> List[TestResultSummarySchema]:
         filtered_test_results_db_rows = self.test_results_db_rows
+        if dbt_invocation and dbt_invocation.invocation_id:
+            filtered_test_results_db_rows = [
+                test_result
+                for test_result in filtered_test_results_db_rows
+                if test_result.invocation_id == dbt_invocation.invocation_id
+            ]
```
haritamar, line 81: I think that if we filter by invocation we should remove the filter below of `invocations_rank_index == 1`, wdyt?
(so we can return results from this invocation even if they are not the last ones)

haritamar (approved):

PR #1858, opened 2025-03-13
Fixed `Example configurations` in the 'Anomaly Tests Params' docs.

nkmr-jp:
Hello! The behavior of the `Example` tab in the `Example configurations` section of the "[Anomaly Tests configuration params](https://docs.elementary-data.com/data-tests/anomaly-detection-configuration/anomaly-params)" documentation was incorrect, so we've fixed it.

When you try to open the second `Example` tab as shown in the video below, the first `Example` tab opens instead, and the second one doesn't open.

The cause is probably that the tab names are duplicated.

https://github.com/user-attachments/assets/a589a0a8-798a-4740-9c28-c0733eaf7664

I modified the tab names as follows, referring to the “[Schema changes from baseline](https://docs.elementary-data.com/data-tests/schema-tests/schema-changes-from-baseline)” page.

• Models
• Models example
• Sources
• Sources example

Now that the tab names are not duplicated, I think this should solve the problem of the tabs not opening.
Also, I added the `version: 2` code, which was missing.

### DIFF
--- a/docs/data-tests/anomaly-detection-configuration/anomaly-params.mdx
+++ b/docs/data-tests/anomaly-detection-configuration/anomaly-params.mdx
@@ -60,7 +60,7 @@ sidebarTitle: "All configuration params"
 
 <CodeGroup>
 
-```yml properties.yml
+```yml Models
 version: 2
 
 models:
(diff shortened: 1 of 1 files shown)


Maayan-s:
Thank you @nkmr-jp !

PR #1866, opened 2025-03-22
use pull_request_target + environment to allow running tests on contributor PRs

haritamar:
…ibutor PRs

### DIFF
--- a/.github/workflows/close_pylon_issue.yml
+++ b/.github/workflows/close_pylon_issue.yml
@@ -3,16 +3,13 @@ name: Close Pylon Ticket on Issue or Pull Request Closure
 on:
   issues:
     types: [closed]
-  pull_request:
+  pull_request_target:
     types: [closed]
 
 jobs:
   close_pylon_ticket:
     runs-on: ubuntu-latest
     steps:
-      - name: Check out the repository
-        uses: actions/checkout@v2
-
       - name: Extract Pylon Ticket ID
         id: extract_ticket_id
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/close_pylon_issue.yml
```diff
@@ -3,16 +3,13 @@ name: Close Pylon Ticket on Issue or Pull Request Closure
 on:
   issues:
     types: [closed]
-  pull_request:
+  pull_request_target:
```
haritamar, line 6: The Pylon actions don't actually need an environment, since they don't need the PR code, just the PR details.
To be safe I actually removed the checkout which is unneeded.

.github/workflows/run-precommit.yml
```diff
@@ -1,14 +1,17 @@
 name: Run pre-commit hooks
 on:
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
```
NoyaArie, line 4: why do we need to change it for the pre-commit?

haritamar: Pre-commit can actually run code from the repo, so I thought it's safer
(you can add scripts to the pre-commit config, or install a bad package via the requirements)

.github/workflows/close_pylon_issue.yml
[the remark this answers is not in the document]
haritamar: Good point, thanks!

.github/workflows/run-precommit.yml
haritamar: Actually we really don't
I thought about that it's running code, but it's not using secrets so no issues here, changing back

NoyaArie (approved):

PR #1887, opened 2025-04-04
Adds --s3-acl option to the cli to be able to set S3 report permissions.

Lawiss:
Currently, when uploading the report to S3, there is no specific permission set, thus it default to `private`.

This PR adds the `--s3-acl` parameter to the CLI in order to be able to set an ACL option (list of possible values are listed [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl)).

This is useful to make the report publicly available when uploaded to the bucket.


michael-myaskovsky (approved):
Thanks for your contribution!

PR #1897, opened 2025-04-15
Use dbt version 1.7.0 as the old version in tests

haritamar:


.github/workflows/remind-docs-and-tests.yml
```diff
@@ -1,6 +1,6 @@
 name: Remind docs and tests
 on:
-  pull_request:
+  pull_request_target:
```
haritamar, line 3: Did this along the way - using `pull_request_target` here since this workflow uses a secret, that will allow it to work for contributor PRs.

ofek1weiss (approved):

PR #1903, opened 2025-05-04
Add function for disable_elementary_logo_print

nickozilla:


michael-myaskovsky:
Hi @nickozilla ! Thanks for your contribution!
The pr fails on pre-commit hooks, if you could fix that, it would be great and then I'll be able to merge.

PR #1913, opened 2025-05-16
Update automated-freshness.mdx to fix a typo Anonmaly

seoruc:
There is typo in the documentation. 
The goal of this PR is to fix that nit...


haritamar:
Thanks for the fix!

PR #1958, opened 2025-07-14
Feature: make the number of columns that are formatted as a table in Teams alerts a cli flag so that users have more control over the formatting in Teams

FrankTub:
I've played around quite a lot with this teams integrations and sadly it is indeed not possible to detect for the new workflows alert webhooks to determine if an alert was sent successfully.

By default the formatting of the tables will behave as it used to do with one exception. The full width is used for the message, independent of the number of columns that the Test Results Sample will contain. So in the old situation:

image

In the new situation it will always look like:

image

When running something like it will simply behave like it used to:

```bash
edr monitor --teams-webhook "$TEAMS_WEBHOOK_URL" --report-url "$DATA_QUALITY_REPORT_URL" --profile-target default
```

The new situation can be triggered with for example:

```bash
edr monitor --teams-webhook "$TEAMS_WEBHOOK_URL" --report-url "$DATA_QUALITY_REPORT_URL" --profile-target default -mc 10
```

Example output:

image

## Summary by CodeRabbit

* **New Features**
  * New config/CLI option to set the maximum columns shown in alert samples; display switches between table and JSON accordingly.
* **Improvements**
  * Adaptive Cards render full-width in Microsoft Teams, wrap long table cell text, and avoid showing "None" for empty cells.
* **Bug Fixes**
  * More reliable Teams webhook handling with clearer success criteria, improved error messages, and rate limiting.
* **Refactor**
  * Reuse a single, configured message builder for consistent alert formatting.
* **Tests**
  * Updated Adaptive Card fixtures for Teams width and text wrapping.
* **Chores**
  * Added ignore rule for internal dbt packages.


elementary/messages/messaging_integrations/teams_webhook.py
```diff
@@ -56,23 +61,31 @@ def __init__(self, url: str) -> None:
...
             raise MessagingIntegrationError(
-                "Failed to send message to Teams webhook"
-            ) from e
+                f"Could not post message to Teams via webhook - {self.url}. Status code: {response.status_code}, Error: {response.text}"
+            )
+        return MessageSendResult(
+            message_context=EmptyMessageContext(),
+            timestamp=datetime.utcnow(),
+            message_format="adaptive_cards",
+        )
```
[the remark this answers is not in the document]
michael-myaskovsky: @FrankTub The second comment here seems relevant, how would the new code handle an exception?

elementary/monitor/alerts/alert_messages/builder.py
```diff
@@ -343,7 +344,8 @@ def _get_result_blocks(
             elif result_sample:
                 if (
                     isinstance(result_sample, list)
-                    and len(result_sample[0].keys()) <= 4
+                    and len(result_sample[0].keys())
+                    <= self.config.maximum_columns_in_alert_samples
                 ):
```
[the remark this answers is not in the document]
michael-myaskovsky: @FrankTub

elementary/monitor/data_monitoring/alerts/data_monitoring_alerts.py
```diff
@@ -309,7 +309,7 @@ def _send_alert(
     ) -> bool:
         if isinstance(self.alerts_integration, BaseIntegration):
             return self.alerts_integration.send_alert(alert)
-        alert_message_builder = AlertMessageBuilder()
+        alert_message_builder = AlertMessageBuilder(self.config)
```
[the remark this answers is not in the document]
michael-myaskovsky: @FrankTub check if the type matches

michael-myaskovsky (commented):
Hi @FrankTub! Thank you for your contribution, I left some comments under coderabbit's comments, if you could take a look and explain/fix I would be glad to approve your contribution.
Thanks again!

elementary/monitor/alerts/alert_messages/builder.py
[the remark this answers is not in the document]
FrankTub: I think this does not makes sense to add the suggestion given that if you look at a bit more of the code that check is already present:

```py
            elif result_sample:
                if (
                    isinstance(result_sample, list)
                    and result_sample
                    and len(result_sample[0].keys())
                    <= self.config.maximum_columns_in_alert_samples
                ):
```

Where this line already in there to solve this:

```py
            elif result_sample:
```

FrankTub:
@michael-myaskovsky I think I addressed all your comments. I will highlight one piece of code which I just don't get how this ever worked in this setup.

elementary/monitor/data_monitoring/alerts/data_monitoring_alerts.py
```diff
@@ -87,6 +90,11 @@ def __init__(
         self.send_test_message_on_success = send_test_message_on_success
         self.override_config_defaults = override_config
         self.alerts_integration = self._get_integration_client()
+        self.alert_message_builder = AlertMessageBuilder(
+            MessageBuilderConfig(
+                maximum_columns_in_alert_samples=self.config.maximum_columns_in_alert_samples
+            )
+        )
```
FrankTub, line 134: Here I created a `MessageBuilderConfig` without the `alert_groups_subscribers` attribute. But in the past I really don't understand if this was ever passed down to the `AlertMessageBuilder`. It seems to me this was only working for the tests, but not in any other way??

michael-myaskovsky:
Great! Please update from main and then I'll run the tests

elementary/messages/formats/adaptive_cards.py
```diff
@@ -131,7 +131,7 @@ def format_fact_list_block(block: FactListBlock) -> Dict[str, Any]:
     }
 
 
-def format_table_block(block: TableBlock) -> Dict[str, Any]:
+def format_table_block(block: TableBlock) -> dict:
```
elazarlachkar: Why this change?

elementary/messages/messaging_integrations/teams_webhook.py
```diff
@@ -42,8 +45,10 @@ def send_adaptive_card(webhook_url: str, card: dict) -> requests.Response:
         headers={"Content-Type": "application/json"},
     )
     response.raise_for_status()
-    if response.status_code == 202:
-        logger.debug("Got 202 response from Teams webhook, assuming success")
+    if response.status_code == HTTPStatus.ACCEPTED:
+        logger.debug(
+            "Got %s response from Teams webhook, assuming success", HTTPStatus.ACCEPTED
```
elazarlachkar: Use `f` formatting here

elementary/messages/messaging_integrations/teams_webhook.py
```diff
@@ -56,22 +61,35 @@ def __init__(self, url: str) -> None:
...
+                )
             return MessageSendResult(
                 message_context=EmptyMessageContext(),
                 timestamp=datetime.utcnow(),
                 message_format="adaptive_cards",
             )
         except requests.RequestException as e:
             raise MessagingIntegrationError(
-                "Failed to send message to Teams webhook"
+                f"Network error while posting message to Teams webhook: {str(e)}"
```
elazarlachkar: RequestException doesn't necessarily mean network error, it could mean invalid JSON, TooManyRedirects etc ([docs](https://requests.readthedocs.io/en/latest/_modules/requests/exceptions/)).

elementary/messages/messaging_integrations/teams_webhook.py
```diff
@@ -56,22 +61,35 @@ def __init__(self, url: str) -> None:
...
         try:
-            send_adaptive_card(self.url, card)
+            response = send_adaptive_card(self.url, card)
+            # The status code is not reliable for the older version and 202 for the newer version
+            # The response returns 1 in the body for the old version for success and otherwise some text
+            # The new version is not reliable to do a check and in the response there is nothing that can be used
+            # to determine if the message was sent successfully.
+            if (
+                response.status_code not in (HTTPStatus.OK, HTTPStatus.ACCEPTED)
+                or len(response.text) > 1
```
elazarlachkar: So if the response has any content, it will be treated as an error?
You wrote in the comment `in the response there is nothing that can be used`, that means there is some content which will always be treated as error, which I don't think we want.

elazarlachkar (changes requested):

elementary/monitor/data_monitoring/alerts/data_monitoring_alerts.py
elazarlachkar: The flag is used by our Cloud product. Your PR keeps the existing implementation in the OSS, so it's Ok :)

elementary/messages/formats/adaptive_cards.py
FrankTub: Probably because in my IDE pylint doesn't like it. Anyways, I made it similar to the rest again

elementary/messages/messaging_integrations/teams_webhook.py
FrankTub: I tried to clarify some things in the comments and in the code. Let me know if it needs more attention. Just to give you some more context:

You have the old webhook version of Teams, in that version they always return status code of `200`. And if there is an error they add it in the response text. If there is no error the response text will be '1'.

The new webhook version of Teams is even worse, there they always return `202` and nothing in the response can be used to determine if it was actually sent.

In cases when you have no internet access or the webhook is down or something another http status code is returned.

elazarlachkar: It's more clear now. Thanks :)

PR #1964, opened 2025-07-18
Fix `float("inf")` and `float("nan")` JSON serialization breaking reports

ClementSicard:
## Description of the problem

- If some failing data tests samples contain some `Infinity` values (possible in BigQuery, this is equivalent to `float("inf")`, then the generated output when [serializing the fetched report data to JSON](https://github.com/elementary-data/elementary/blob/master/elementary/monitor/data_monitoring/report/data_monitoring_report.py#L77), then dumping it to `elementary_output.json` contains `Infinity` values, since `allow_nan` flag to `json.dumps` is `True` by default (see [here](https://docs.python.org/3/library/json.html#json.dump:~:text=Default%20True.-,allow_nan,-(bool)%20%E2%80%93%20If)).
- The result is that the HTML report cannot interpret `Infinity` in the JSON, as it [doesn't comply with the JSON specification](https://datatracker.ietf.org/doc/html/rfc8259#:~:text=Numeric%20values%20that%20cannot%20be%20represented%20in%20the%20grammar%20below%20(such%0A%20%20%20as%20Infinity%20and%20NaN)%20are%20not%20permitted.)
- This leads to the generated HTML report appearing as empty because the embedded JSON cannot be successfully decoded.

## Resolution

- This PR introduces `inf_and_nan_to_str`, so that `float("inf")` gets turned into `"Infinity"` and `float("nan")` into `"NaN"` as a `json_utils` method (hopefully that is an appropriate location).
- The method is called right before the `json.dumps` call.

> [!NOTE] 
> Note that we couldn't simply use a `default=` or `cls=` method in the dumps arg, something like
> ```python
> class CustomEncoder(json.JSONEncoder):
>     def default(self, obj):
>         if isinstance(obj, float):
>             if math.isinf(obj):
>                 return "Infinity" if obj > 0 else "-Infinity"
>             elif math.isnan(obj):
>                 return "NaN"
>         return super().default(obj)
> 
> json_str = json.dumps(data, cls=CustomEncoder)
>```
> because the `json` modules intercepts these values and turns them into unquoted strings _before_ the custom encoder can be applied...

## Summary by CodeRabbit

* **New Features**
  * Improved handling of special float values (Infinity, -Infinity, NaN) in data monitoring reports by converting them to string representations, ensuring successful JSON serialization.

* **Tests**
  * Added a test to verify conversion of special float values to strings before JSON serialization.


arbiv (approved):

arbiv:
Thanks @ClementSicard! Merging :)

PR #1984, opened 2025-08-15
Slack join channel recursion fix

MikaKerman:
Added an option to only list public channels, since those are the ones we can join.

## Summary by CodeRabbit

- New Features
  - Improved Slack channel discovery with an option to limit lookups to public channels, enhancing reliability for public-channel messaging.
- Bug Fixes
  - Reduced message failures when the bot isn’t in a channel by prioritizing public-channel resolution and joining when possible.
  - Avoids unnecessary attempts to access private channels, leading to fewer errors and faster delivery in public channels.
- Improvements
  - More accurate channel matching and pagination handling for Slack channel lists, improving overall messaging stability.

### DIFF
--- a/elementary/messages/messaging_integrations/slack_web.py
+++ b/elementary/messages/messaging_integrations/slack_web.py
@@ -114,7 +114,7 @@ class SlackWebMessagingIntegration(
             logger.info(
                 f'Elementary app is not in the channel "{channel_name}". Attempting to join.'
             )
-            channel_id = self._get_channel_id(channel_name)
+            channel_id = self._get_channel_id(channel_name, only_public=True)
             self._join_channel(channel_id=channel_id)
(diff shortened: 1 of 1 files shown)


elementary/messages/messaging_integrations/slack_web.py
```diff
@@ -114,7 +114,7 @@ def _handle_send_err(self, err: SlackApiError, channel_name: str):
             logger.info(
                 f'Elementary app is not in the channel "{channel_name}". Attempting to join.'
             )
-            channel_id = self._get_channel_id(channel_name)
+            channel_id = self._get_channel_id(channel_name, only_public=False)
```
oravi: Here it should be only_public=True no?

oravi (approved):

Issue #1986, opened 2025-08-19
Use ' instead of " to wrap database_and_schema in Slack for "The test query was too long, here's a query to get it."


mschmidoev:
**Describe the bug**
When "The test query was too long, here's a query to get it." runs it posts the database_and_schema with quotation `"` marks around it. This causes the query not to run in SQL query editors like Datagrip or Databricks as it's incorrect syntax. 

Currently it's returning this in Slack
**Test query**
_The test query was too long, here's a query to get it._
```
SELECT test_results_query FROM "database.schema".elementary_test_results WHERE test_execution_id = '1234.test.xyz.test_model_name'
```

**Expected behavior**
Would be better if it returned it wrapped in `'`s so it could run straight from copying and pasting without any editing needed.
So ideally it would return this:
**Test query**
_The test query was too long, here's a query to get it._
```
SELECT test_results_query FROM 'database.schema'.elementary_test_results WHERE test_execution_id = '1234.test.xyz.test_model_name'
```

**Screenshots**

**Would you be willing to contribute a fix for this issue?**
Yes, will post PR shortly

NoyaOffer:
Thanks @mschmidoev ! We'll have a look.

elazarlachkar:
Hi @mschmidoev!

If I understand correctly, the quoting rules depend on the SQL dialect being used.
Databases that follow Standard SQL (ANSI), like PostgresQL, use double quotes for identifiers.
Other databases may use backticks (MySQL) or even square brackets (SQL server).

If this is indeed the case, I would prefer to stick with Standard SQL.
What's your thoughts on it?

mschmid09:
Thanks @elazarlachkar  and @NoyaOffer.
Ah sorry, I think I wasn't very clear, and I think I need to update my PR and description which I'll do in a sec.
The issue isn't the identifiers actually, it's because it's joining the database and schema together like  `"database.schema"` which treats the whole thing as one identifier. Would be best to split it into `"database"."schema"`

elazarlachkar:
@mschmidoev Got you, that's make sense :)
I approved the PR, thanks for the contribution!

mschmid09:
Thanks @elazarlachkar, looking forward to this change getting merged. Let me know if there's anything I still need to do :)

PR #1987, opened 2025-08-19
Fix: Replace double quotes with single quotes in elementary_database_and_schema

mschmidoev:
- Replace double quotes with single quotes for proper SQL compatibility
- Fixes SQL errors in Slack alert messages when database/schema names contain quotes

https://github.com/elementary-data/elementary/issues/1986


elazarlachkar:
@mschmidoev The CI is failing for some reason. Can you update your branch to the latest `master`?

mschmid09:
@elazarlachkar should be good to go :)

elazarlachkar:
Hi @mschmidoev!

The CI still has the issue we solved on `master`.
Can you make sure you're updated to the latest `elementary-data:master`? I assume you updated from the fork's master (`mschmidoev:master`).

Thanks, and sorry for the trouble.

mschmid09:
Hey @elazarlachkar , no dramas, should be good now

elazarlachkar:
Hi @mschmidoev!
The CI issue is now fixed on the latest `elementary-data:master`. Sorry for the trouble, can you update your branch again?

mschmid09:
Thanks @elazarlachkar, was off last week but this PR should be up to date with master branch now!

mschmid09:
@elazarlachkar looks like that's worked this time around! Let me know if there's anything else you need :) )

mschmid09:
Hey @elazarlachkar just checking if there's any progress here :)

elementary/monitor/data_monitoring/data_monitoring.py
```diff
@@ -79,6 +79,12 @@ def get_elementary_database_and_schema(self):
             relation = self.internal_dbt_runner.run_operation(
                 "elementary_cli.get_elementary_database_and_schema", quiet=True
             )[0]
+            # Split on dot and wrap each part in double quotes for standard SQL compatibility
```
elazarlachkar: Hi @mschmidoev!

2 things we noticed here:

1. The `internal_dbt_runner` returns the `relation` object **already** wrapped with double quotes (`"my_db.my_schema"`). That's the issue you encountered.  
That means that by splitting the `.` - you'll get 2 parts, each part with a single `"`. db will be `"my_db` and schema will be `my_schema"`.  
Then wrapping each part with `"` and merging them will result in `""my_db"."my_schema""`, which have an extra `"`.

2. For Snowflake, the table identifiers are case insensitive by default. Wrapping the identifier with `"` changes that to be case sensitive, which might cause issues.  
This is something unrelated to your change of course.

For these 2 reason, I suggest to just use the `relation` returned from the `internal_dbt_runner`, and strip the `"`. quotes from it entirely. So instead of `"my_db.my_schema"` - using `my_db.my_schema`. What do you think?

```
relation = self.internal_dbt_runner.run_operation(
      "elementary_cli.get_elementary_database_and_schema", quiet=True
)[0]
strip_relation = relation.strip('"')
```

mschmid09: Hey @elazarlachkar, no worries and sorry for my delayed reply, just returned from AL.
Yep That seems to be a cleaner solution all round, have committed that change :)

elazarlachkar: Thanks @mschmidoev!

elazarlachkar:
@mschmidoev I merged the change, thanks for the PR!
It will be included in the next release (should be in the next few days).

PR #2005, opened 2025-09-09
Handle invalid characters when uploading CI artifact

elazarlachkar:
## Description

When branch names include `/`, the CI fails when trying to upload an artifact:
```
Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n, Backslash \, Forward slash /
```

See [example](https://github.com/elementary-data/elementary/actions/runs/17451278353/job/49556212658):
```
Error: The artifact name is not valid: report_postgres_fix/replace-double-quotes-with-single-quotes_dbt_.html. Contains the following character:  Forward slash /
```

This PR fixes it by replacing all the invalid characters with `_`.


NoyaArie (approved):
👍🏼

PR #2019, opened 2025-10-01
change docs to match dbt fusion

haritamar:


.pre-commit-config.yaml
```diff
@@ -44,8 +44,11 @@ repos:
         language: system
         files: ^elementary/.*\.py$
 
-  - repo: https://github.com/CoderJoshDK/precommit-mintlify-validate/
```
haritamar, line 47: This is a validation that we have no broken links in mintlify.
It didn't really run lately as docs aren't often updated by devs anymore - but it is useful so I made it work.

PR #2039, opened 2025-10-28
Ele 5146 bugfix to disable_samples flag + update package ref

haritamar:
null

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of test result sample/metrics data: disables sample payloads when configured, consistently uses metrics for anomaly detection, and stabilizes sorting and type handling to produce more reliable test results.

* **Chores**
  * Updated dependency sourcing and lock metadata to pin a specific data-reliability revision and adjusted package declarations to reflect the new source.

### DIFF
--- a/elementary/monitor/api/tests/tests.py
+++ b/elementary/monitor/api/tests/tests.py
@@ -448,7 +448,7 @@ class TestsAPI(APIClient):
             # Sample data is always a list for non-elementary tests
             sample_data = cast(Optional[list], test_result_db_row.sample_data)
             if disable_samples:
-                sample_data = []
+                sample_data = None
 
             test_results = DbtTestResultSchema(
                 display_name=test_result_db_row.test_name,


elementary/monitor/api/tests/tests.py
```diff
@@ -444,10 +444,11 @@ def _get_test_result_from_test_result_db_row(
     ) -> Optional[Union[DbtTestResultSchema, ElementaryTestResultSchema]]:
         test_results: Optional[Union[DbtTestResultSchema, ElementaryTestResultSchema]]
 
-        sample_data = test_result_db_row.sample_data if not disable_samples else None
```
haritamar, line 447: There was a bug here where the `disable_samples` flags also removed anomaly tests rows (metrics), which is incorrect

elementary/monitor/api/tests/tests.py
```diff
@@ -444,10 +444,11 @@ def _get_test_result_from_test_result_db_row(
...
         if test_result_db_row.test_type == "dbt_test":
             # Sample data is always a list for non-elementary tests
-            sample_data = cast(Optional[list], sample_data)
+            sample_data = cast(Optional[list], test_result_db_row.sample_data)
+            if disable_samples:
+                sample_data = []
```
elazarlachkar: Consider adding a log

haritamar: this function is called per line, so I think it might be a bit spamming

PR #2043, opened 2025-10-30
Correctly shorten versioned model names

Ryosuke839:
Elementary shortens model name in Slack alerts. However, it will only show model version like `v2` when tests were against versioned models https://docs.getdbt.com/docs/mesh/govern/model-versions.
(Note: Slack alert message is generated in https://github.com/elementary-data/elementary/blob/ba986eb94b942cb8312545d6100387c71c68cb85/elementary/monitor/alerts/alerts_groups/grouped_by_table.py#L24-L40)

image

This PR changes shortening logic so that shortened model name will consist of both model base name and model version like `model.v2`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shortening of versioned model names so both the model identifier and its version are displayed when applicable (e.g., four-part names now show the last two segments).

* **Tests**
  * Added unit tests validating model-name shortening across edge cases: None, single-part, three-part, and four-part inputs.

### DIFF
--- a/tests/unit/utils/test_models.py
+++ b/tests/unit/utils/test_models.py
@@ -16,6 +16,6 @@ def test_shorten_3_parts():
     assert shortened == "bar"
 
 
-def test_shorten_1_part():
+def test_shorten_4_parts():
     shortened = get_shortened_model_name("models.foo.bar.v2")
     assert shortened == "bar.v2"


NoyaOffer:
Hey @Ryosuke839 ! Thank you for your contribution, appreciate it.
We're taking a look and hope to approve soon 😄

NoyaOffer:
@Ryosuke839 Thanks again for the contribution!
We're adding a few more fixes and features before the next version release, you can use Master for now 😄

PR #2093, opened 2026-01-21
Make default Config() Initialization respect target_path in default config file

ryanquincypaul:
### The problem

In databricks python tasks using serverless compute, the default current working directory doesn't allow writes. The logo and version checking code instantiates a Config object with the default values and no way to configure them which means target_path goes to `DEFAULT_TARGET_PATH = os.getcwd() + "/edr_target"` and no way to override it. Below the solution I will put lines of code of the issues in case that is relevant.

### What I would like to have happen

I don't think the logo and version checking logic should require file writes that I have no option of skipping at all, but in lieu of that, instead of `os.chdir()`-ing the working directory to somewhere like `/tmp`, I'd rather just use the already existing target_path option to tell edr to put files in a place where I am allowed to write. The logo and upgrade check calls use a default Config() option which forces a default path using os.getcwd(). I would like it so at least options in a config file written at the default location (`~/.edr/config.yml`) is respected in a default `Config()` call. I can then at least give it a `target-path: /tmp/WHATEVER_{unique_timestamp}` or something so it can create the `edr_target` directory just to print the logo and check the dbt and python package version numbers in peace. It would be great if it would also respect the arguments I am passing in as I am using the --target-path parameter for `edr report` to store it somewhere in a databricks volume.

### What I did

I am just setting the default value of target_path to None so that it gets to the check for a value in the default config directory in the `_first_not_none` set, and then it will go to the original DEFAULT_TARGET_PATH. I am also setting the DBT_LOG_PATH env var to the target_dir and not the passed in target_path.

- New default:

https://github.com/ryanquincypaul/elementary/blob/8243a7068b21a72ad0bbc921b5f04c0bca6131b3/elementary/config/config.py#L49

- New value checking and DBT_LOG_PATH setting code:

https://github.com/ryanquincypaul/elementary/blob/8243a7068b21a72ad0bbc921b5f04c0bca6131b3/elementary/config/config.py#L93-L100

### Other solutions I had thought of
- Make sure logo and version checking run after arguments have been parsed so they can be used OR after a Config object has already been built in a way that respects cli arguments
  - This would still require changes to the config code because target_path argument is used as is for the purposes of setting DBT_LOG_PATH
  - Also that would defeat the purpose of these being simple things that always run at the top regardless if they require file writes lol.
- Just add more environment variables that can be used to skip the whole logo and version calls OR to manually override that DEFAULT_TARGET_PATH in Config class
  - A little too nuclear and hacky. Also I don't see a wide usage of environment variable overrides in the code base so not really fitting the paradigm.

### More details on issue
Should be noted that there were PRs made and accepted adding configuration to disable the logo and version checking:

https://github.com/elementary-data/elementary/pull/1903
https://github.com/elementary-data/elementary/pull/1674

- Here are calls to print the logo and do the dbt vs python package version checking at the top of the cli file.

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/cli/cli.py#L15-L16

- There are PRs that added options skip them but because they both use `Config()` with no parameters given or any other way to override the target_path. 

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/cli/logo.py#L15-L21

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/cli/upgrade.py#L10-L14

- In the Config code, we can see the following default behavior:
  - Defaults for config and target_path

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/config/config.py#L36-L38

  - Default arguments

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/config/config.py#L44

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/config/config.py#L49

  - The check for target_path that has no way of getting to the config file since the default will always be present and I'm not sure you can even pass `None` via the command line if these default Config() calls even listened to that input. Furthermore, the DBT_LOG_PATH setting doesn't even respect the config path anyway, it's only using what's passed into the `target_path` parameter which again is just the default one in logo and variable checking code.

https://github.com/elementary-data/elementary/blob/398b34a99fa699d7a7c7a22d70e486fa86eda9d6/elementary/config/config.py#L93-L99

## Summary by CodeRabbit

* **Refactor**
  * Made the target path optional and adjusted fallback to use the default target directory when unset.
  * Logging now uses the resolved target directory instead of the original input.
  * Directory creation behavior continues to rely on the resolved target directory.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/elementary/cli/cli.py
+++ b/elementary/cli/cli.py
@@ -29,6 +29,13 @@ def get_log_path(ctx):
         return os.path.join(target_path, "edr.log")
 
 
+def get_quiet_logs(ctx):
+    try:
+        return "--quiet-logs" in ctx.args
+    except (ValueError, AttributeError):
+        return False
+
+
 class ElementaryCLI(click.MultiCommand):
     _CMD_MAP = {
         "monitor": monitor,
@@ -57,12 +64,14 @@ class ElementaryCLI(click.MultiCommand):
 
     def invoke(self, ctx: click.Context) -> Any:
         files_target_path = get_log_path(ctx)
-        set_root_logger_handlers("elementary", files_target_path)
-        click.echo(
-            "Any feedback and suggestions are welcomed! join our community here - "
-            "https://bit.ly/slack-elementary\n"
-        )
-        logger.info(f"Running with edr={package.get_package_version()}")
+        quiet_logs = get_quiet_logs(ctx)
+        set_root_logger_handlers("elementary", files_target_path, quiet_logs=quiet_logs)
+        if not quiet_logs:
+            click.echo(
+                "Any feedback and suggestions are welcomed! join our community here - "
+                "https://bit.ly/slack-elementary\n"
+            )
+            logger.info(f"Running with edr={package.get_package_version()}")
         return super().invoke(ctx)
 
 
--- a/elementary/config/config.py
+++ b/elementary/config/config.py
@@ -75,6 +75,7 @@ class Config:
         env: str = DEFAULT_ENV,
         run_dbt_deps_if_needed: Optional[bool] = None,
         project_name: Optional[str] = None,
+        quiet_logs: Optional[bool] = None,
     ):
         self.config_dir = config_dir
         self.profiles_dir = profiles_dir
@@ -216,6 +217,10 @@ class Config:
 
         self.disable_elementary_logo_print = config.get(
             "disable_elementary_logo_print", False
+        )
+
+        self.quiet_logs = self._first_not_none(
+            quiet_logs, config.get("quiet_logs"), False
         )
 
     def _load_configuration(self) -> dict:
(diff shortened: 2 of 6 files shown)


ryanquincypaul:
I did not mean to make this to the main repo, just my fork as I was having trouble with local unit tests even with unmodified code. I was just pushing to see the unit tests run in the github actions as I couldn't get them to pass locally. I was just going to build a wheel from this and use it manually in databricks tasks! But whatever it is out there now. I will make a little more effort in the description.

ryanquincypaul:
To be clear, i'm not requesting anything. It would be cool if this was builtin, but as of now I am just going to build a wheel and try it out in databricks now and I can report back. I didn't mean for this to be visible lol. Open to any feedback though regardless.

ryanquincypaul:
@coderabbitai summary

ryanquincypaul:
So this worked! But adding `target-path: /tmp/whatever` and staying in the default current working directory and providing the actual directory path for the profiles-dir and project-dir arguments still didn't work because it couldn't write a package-lock.json file when running the actual command. So I end up just changing the directory anyway. I'll keep this open as I continue to work on this and set a reminder to check in on friday and close this if there is no change.

PR #2107, opened 2026-02-10
feat: add --slack-full-width for full-width Slack alerts with markdown test result tables

### LINKED ISSUE #1079: Formatting the test_results_sample as a table within the Slack alters
When generating the Slack Alerts, currently the test_results_sample returns a JSON value. It would be much nicer to have this formatted as a table.

The test_results_sample is formatted like a table. 

**Additional context**
This is a feature that would be greatly appreciated by the consumers of these slack alerts who are not always technically inclined.

michrzan:
## Summary

This PR adds a --slack-full-width option that allows Slack alerts to use the full message width by rendering the entire message with Block Kit (no attachments) and displaying test result samples as markdown tables.

This significantly improves readability when test results contain many columns or long values.

### Problem

When the Test Results Sample includes multiple columns or long strings, the current Slack alerts become difficult to read due to the narrow attachment layout. In practice, this makes it hard for alert recipients to quickly understand the context and take action.

old

or when there is more columns like this:

old_big

### Solution

With the --slack-full-width option enabled, Slack alerts are rendered at full width and test results are consistently shown as markdown tables:

new

or when there is more columns like this:

new_big

Admittedly, it looks better on a wide screen.

### Implementation details

This is achieved by:
- Building the Slack message without attachments and rendering everything using Block Kit blocks
- Inserting an empty rich text block at the start of the message to force full-width rendering
- Always rendering test results as markdown tables

### Trade-offs

As a result of moving away from attachments, the colored severity indicator (yellow/red pipe on the left) is no longer available. This is a known limitation of the approach.

### Notes

This resolves #1079.

This is jsut something I am using internally and wanted to share. Happy to discuss alternatives, trade-offs, or refinements!

## Summary by CodeRabbit

* **New Features**
  * Added a --slack-full-width CLI flag to enable a full-width Slack alert layout.

* **Improvements**
  * Slack alerts: preview/details render inline for full-width mode, attachments suppressed, and titles include a “Powered by Elementary” section; headers/icons and description/column handling were refined.
  * Table rendering: numeric-like values are treated as text and markdown table conversion added for result samples.

* **Tests**
  * Added tests for full-width Slack behavior and markdown table formatting.

### DIFF
--- a/docs/oss/deployment-and-configuration/slack.mdx
+++ b/docs/oss/deployment-and-configuration/slack.mdx
@@ -53,15 +53,3 @@ The alert format is:
 ```
 
 ---
-
-## Full-width alerts
-
-By default, Slack alerts use a narrower layout with some content in attachments. To use the full message width and show test results as a Markdown table in the main message body, pass the flag when running the monitor:
-
-```shell
-edr monitor --slack-token <your_slack_token> --slack-channel-name <channel> --slack-full-width
-```
-
-With `--slack-full-width`, alerts are sent using Slack Block Kit in the main message body instead of attachments, so the full channel width is used and test result samples appear as formatted Markdown tables.
-
----
--- a/docs/oss/deployment-and-configuration/slack.mdx
+++ b/docs/oss/deployment-and-configuration/slack.mdx
@@ -53,3 +53,15 @@ The alert format is:
 ```
 
 ---
+
+## Full-width alerts
+
+By default, Slack alerts use a narrower layout with some content in attachments. To use the full message width and show test results as a markdown table in the main message body, pass the flag when running the monitor:
+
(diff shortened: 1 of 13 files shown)


michrzan:
Thanks for the commits @haritamar 

I rebased the branch to master. Tested locally and all seems to be working fine.

I also aligned `_get_elementary_test_template` description rendering with `_get_dbt_test_template` (single combined section block instead of separate header + context block) as per one of the nitpicks.

NumberPiOso:
Hi !  Yes, my workaround has been to always show 5 columns so it avoids truncation since it automatically changes it to JSON.
I think it would be desireable to have the output format to JSON directly? and assume that JSON will not be truncated.

PR #2164, opened 2026-03-21
Teams webhook message too large exception

GuyEshdat:
null

## Summary by CodeRabbit

* **New Features**
  * Teams webhook integration now surfaces a specific "payload too large" error for oversized adaptive card payloads.

* **Bug Fixes**
  * Removed automatic/truncated sending of oversized Teams adaptive cards; when payloads exceed limits, users receive an explicit error instead of a silently truncated message.

### DIFF
--- a/elementary/messages/messaging_integrations/teams_webhook.py
+++ b/elementary/messages/messaging_integrations/teams_webhook.py
@@ -1,7 +1,6 @@
-import json
 from datetime import datetime, timezone
 from http import HTTPStatus
-from typing import Any, Dict, List, Optional
+from typing import Any, Optional
 
 import requests
 from ratelimit import limits, sleep_and_retry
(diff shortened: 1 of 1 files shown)


ofek1weiss (approved):

WillC3:
Hi @GuyEshdat / @ofek1weiss I was just wondering why you reverted this change? I'm looking to do similar for slack, since we are hitting errors when the payload is too large. But obviously won't if you don't think it's a good idea..

I did try a fix of adding `alert_fields` to the test so that it didn't add the query (which was long due to the long list of accepted values) to the payload, but that doesn't seem to work.

```
- accepted_values:
              name: accepted_manufacturer_codes
              meta:
                alert_fields: ["table", "column", "description", "owners", "tags", "subscribers", "result_message"]
              values:
                - ACV
                - ADISA
                ...
```

PR #2196, opened 2026-04-22
perf: cache parsed manifest in APIDbtRunner

dtaniwaki:
Hi, thanks for building and maintaining Elementary! I'd love to get your review on this small performance improvement.

### Problem

`edr report` executes ~18 sequential `dbt run-operation` calls via the Python API. Each call creates a fresh `dbtRunner()` without passing a manifest, which forces a full manifest re-parse every time. On a project with hundreds of models, sources, and macros, each parse can take ~10 seconds — adding up to **~3 minutes of overhead** while the actual queries finish in seconds.

### Solution

Cache the parsed manifest on `APIDbtRunner` after the first successful invocation and reuse it for all subsequent `dbtRunner(manifest=...)` calls. This brings the number of manifest parses from ~18 down to 1.

Key details:
- `self._manifest` is initialized before `super().__init__()` because the parent constructor may call `_inner_run_command()` via `_run_deps_if_needed()`.
- The manifest is only cached after a successful invocation to avoid storing a partial state.
- `dbtRunner.manifest` is a public API available since dbt-core 1.5+.
- Only `APIDbtRunner` is affected — `SubprocessDbtRunner` and `DbtFusionRunner` are unchanged.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced dbt command execution to cache and reuse manifest data across multiple command invocations for improved efficiency.

### DIFF
--- a/elementary/clients/dbt/api_dbt_runner.py
+++ b/elementary/clients/dbt/api_dbt_runner.py
@@ -24,6 +24,10 @@ class APIDbtCommandResult(DbtCommandResult):
 
 
 class APIDbtRunner(CommandLineDbtRunner):
+    def __init__(self, *args, **kwargs):
+        self._manifest = None
+        super().__init__(*args, **kwargs)
+
     def _inner_run_command(
         self,
         dbt_command_args: List[str],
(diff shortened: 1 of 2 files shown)


dtaniwaki:
Closing in favor of #2197 which should address this more effectively.

PR #2197, opened 2026-04-22
feat(report): add --threads option to fetch report data in parallel

dtaniwaki:
Hi maintainers, thanks for the great work on Elementary! I'd appreciate a review when you get a chance.

## Problem

Generating the Elementary report can be slow on larger projects because `ReportAPI.get_report_data` issues a long sequence of `dbt run-operation` calls (models, sources, seeds, snapshots, tests, test results, source freshness, invocations, coverages, lineage, …) one after another. Each call has a non-trivial dbt startup cost on top of the actual query time, so end-to-end report generation is bottlenecked on serial fetch latency even though most of these fetches are independent of each other.

Two things make this tricky to parallelize today:

1. The default in-process `APIDbtRunner` relies on dbt globals (`GLOBAL_FLAGS`, the adapter `FACTORY`, etc.) that are not safe to drive from multiple threads.
2. The existing flow interleaves fetching and post-processing, so it isn't obvious which calls can actually run concurrently.

## Solution

This PR adds an opt-in `--threads N` option (default `1`, preserving current behavior) to both `edr report` and `edr send-report`.

When `threads > 1`:

- A `SubprocessDbtRunner` is used for the parallel fetches. Each call spawns an independent dbt process, sidestepping the in-process global state and making it safe to drive concurrently from a `ThreadPoolExecutor`.
- `get_report_data` is split into phases by data dependency:
  - **Phase 1** — fully independent fetches (models, sources, seeds, snapshots, tests, test/source-freshness results & runs, coverages, invocations, …) run in parallel.
  - **Phase 2** — fetches that need Phase 1 IDs (`exposures` from upstream node IDs, `test_results` keyed by the test invocation) run in parallel.
  - **Phase 3** — `lineage` is fetched once Phase 2 node IDs are known.
  - **Phase 4** — pure in-memory assembly (no dbt calls), shared with the sequential path via a new `_assemble_report_data` helper.

```mermaid
flowchart TD
    Start([edr report --threads N]) --> P1
    subgraph P1[Phase 1: independent fetches in parallel]
        direction LR
        M[models] ~~~ S[sources] ~~~ Se[seeds] ~~~ Sn[snapshots] ~~~ T[tests] ~~~ TR[test runs] ~~~ SFR[source freshness runs] ~~~ C[coverages] ~~~ I[invocations]
    end
    P1 --> P2
    subgraph P2[Phase 2: needs Phase 1 IDs, in parallel]
        direction LR
        E[exposures] ~~~ TRes[test_results]
    end
    P2 --> P3
    subgraph P3[Phase 3: needs Phase 2 node IDs]
        L[lineage]
    end
    P3 --> P4[Phase 4: in-memory assembly]
    P4 --> End([report data])
```

The sequential path is unchanged when `--threads 1` (the default), so users who don't opt in see no behavior change.

Example:

```bash
edr report --threads 8
edr send-report --slack-token ... --threads 8
```

A unit test (`tests/unit/monitor/api/report/test_report_parallel.py`) covers the parallel orchestration and confirms the assembled output matches the sequential path.

Happy to adjust the option name, default, or factor things differently if you'd prefer a different shape.

### DIFF
--- a/elementary/monitor/api/report/report.py
+++ b/elementary/monitor/api/report/report.py
@@ -1,7 +1,9 @@
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 from elementary.clients.api.api_client import APIClient
+from elementary.clients.dbt.subprocess_dbt_runner import SubprocessDbtRunner
 from elementary.monitor.api.filters.filters import FiltersAPI
 from elementary.monitor.api.groups.groups import GroupsAPI
 from elementary.monitor.api.groups.schema import GroupsSchema
@@ -38,7 +40,10 @@ from elementary.monitor.api.tests.tests import TestsAPI
 from elementary.monitor.api.totals_schema import TotalsSchema
 from elementary.monitor.data_monitoring.schema import SelectorFilterSchema
 from elementary.monitor.fetchers.tests.schema import NormalizedTestSchema
+from elementary.utils.log import get_logger
 from elementary.utils.time import get_now_utc_iso_format
+
+logger = get_logger(__name__)
 
 
 class ReportAPI(APIClient):
@@ -68,7 +73,66 @@ class ReportAPI(APIClient):
     ) -> Dict[str, NormalizedExposureSchema]:
         return models_api.get_exposures(upstream_node_ids=upstream_node_ids)
 
+    def _create_subprocess_runner(self) -> SubprocessDbtRunner:
+        """Create a SubprocessDbtRunner for thread-safe parallel execution.
+
+        dbt's Python API (APIDbtRunner) is not thread-safe due to global
+        mutable state (GLOBAL_FLAGS, adapter FACTORY, etc.).
+        SubprocessDbtRunner spawns an independent dbt process per call,
+        making it safe to use from multiple threads.
+        """
+        runner = self.dbt_runner
+        return SubprocessDbtRunner(
+            project_dir=runner.project_dir,
+            profiles_dir=runner.profiles_dir,
+            target=runner.target,
+            raise_on_failure=runner.raise_on_failure,
+            env_vars=getattr(runner, "env_vars", None),
+            vars=runner.vars,
+            secret_vars=runner.secret_vars,
+            allow_macros_without_package_prefix=runner.allow_macros_without_package_prefix,
+            run_deps_if_needed=False,
+        )
+
     def get_report_data(
+        self,
+        days_back: int = 7,
+        test_runs_amount: int = 720,
+        disable_passed_test_metrics: bool = False,
+        exclude_elementary_models: bool = False,
+        project_name: Optional[str] = None,
+        disable_samples: bool = False,
+        filter: SelectorFilterSchema = SelectorFilterSchema(),
+        env: Optional[str] = None,
+        warehouse_type: Optional[str] = None,
+        threads: int = 1,
+    ) -> Tuple[ReportDataSchema, Optional[Exception]]:
+        if threads > 1:
+            return self._get_report_data_parallel(
+                days_back=days_back,
+                test_runs_amount=test_runs_amount,
+                disable_passed_test_metrics=disable_passed_test_metrics,
+                exclude_elementary_models=exclude_elementary_models,
+                project_name=project_name,
+                disable_samples=disable_samples,
+                filter=filter,
+                env=env,
+                warehouse_type=warehouse_type,
+                threads=threads,
+            )
+        return self._get_report_data_sequential(
+            days_back=days_back,
+            test_runs_amount=test_runs_amount,
+            disable_passed_test_metrics=disable_passed_test_metrics,
+            exclude_elementary_models=exclude_elementary_models,
+            project_name=project_name,
+            disable_samples=disable_samples,
+            filter=filter,
+            env=env,
+            warehouse_type=warehouse_type,
+        )
+
+    def _get_report_data_sequential(
         self,
         days_back: int = 7,
         test_runs_amount: int = 720,
@@ -112,15 +176,6 @@ class ReportAPI(APIClient):
             lineage_node_ids.extend(exposures.keys())
             singular_tests = tests_api.get_singular_tests()
 
-            groups = self._get_groups(
-                models.values(),
-                sources.values(),
-                exposures.values(),
-                seeds.values(),
-                snapshots.values(),
-                singular_tests,
-            )
-
             models_runs = models_api.get_models_runs(
                 days_back=days_back, exclude_elementary_models=exclude_elementary_models
             )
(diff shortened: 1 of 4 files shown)


elementary/monitor/api/report/report.py
```diff
@@ -68,6 +73,27 @@ def _get_exposures(
...
+            project_dir=runner.project_dir,
+            profiles_dir=runner.profiles_dir,
+            target=runner.target,
+            raise_on_failure=runner.raise_on_failure,
+            env_vars=getattr(runner, "env_vars", None),
+            vars=runner.vars,
+            secret_vars=runner.secret_vars,
+            allow_macros_without_package_prefix=runner.allow_macros_without_package_prefix,
+            run_deps_if_needed=False,
+        )
```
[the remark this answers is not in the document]
dtaniwaki: Thanks for flagging this. Considered it but leaving as-is for this PR:

- `run-operation` is read-only from elementary's perspective — we only read the macro output from stdout. `run_results.json` written by dbt is never consumed by elementary, so concurrent overwrites in `target/` do not affect correctness.
- `partial_parse.msgpack` clobbers only invalidate dbt's next-startup parse cache; the current run still parses successfully.
- dbt log writes are POSIX appends, so interleaving is at worst cosmetic.

Per-worker `--target-path` / `--log-path` isolation would require extending `SubprocessDbtRunner` to inject CLI flags per call, which is out of scope for this PR. Happy to revisit in a follow-up if maintainers want hard isolation.

elementary/monitor/cli.py
```diff
@@ -449,6 +449,13 @@ def monitor(
     default=True,
     help="Whether to open the report in the browser.",
 )
+@click.option(
+    "--threads",
+    type=int,
+    default=1,
+    help="Number of threads for fetching report data in parallel. "
+    "When set to >1, independent dbt operations run concurrently using subprocess-based runners.",
+)
```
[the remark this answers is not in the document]
dtaniwaki: Already addressed in 3ed15b9 — `--threads` on both `report` and `send-report` now uses `click.IntRange(min=1)` with `show_default=True`.

tests/unit/monitor/api/report/test_report_parallel.py
```diff
@@ -0,0 +1,121 @@
...
+            mock_pool_cls.assert_called_with(max_workers=4)
+
+    def test_error_propagation(self, mock_dbt_runner):
+        api = ReportAPI(mock_dbt_runner)
+        error = RuntimeError("test error")
+        with patch.object(
+            api, "_create_subprocess_runner", side_effect=error
+        ):
+            result, err = api._get_report_data_parallel(threads=4)
+            assert err is error
```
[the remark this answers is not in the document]
dtaniwaki: Already addressed in 6048e93 — the test now returns a mock with `invocation_id`, asserts `err is None`, asserts `_assemble_report_data` was called, and renames the unused `result` to `_`.

tests/unit/monitor/api/report/test_report_parallel.py
```diff
@@ -0,0 +1,125 @@
...
+import pytest
+
+from elementary.monitor.api.report.report import ReportAPI
+
+
+@pytest.fixture
+def mock_dbt_runner():
+    runner = MagicMock()
+    runner.project_dir = "/tmp/project"
+    runner.profiles_dir = "/tmp/profiles"
```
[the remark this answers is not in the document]
dtaniwaki: Addressed in 7b2df2b — `/tmp/project` and `/tmp/profiles` replaced with `mock/project` and `mock/profiles` (Ruff S108 no longer triggers).

elementary/monitor/api/report/report.py
```diff
@@ -136,98 +191,278 @@ def get_report_data(
...
+        days_back: int = 7,
+        test_runs_amount: int = 720,
+        disable_passed_test_metrics: bool = False,
+        exclude_elementary_models: bool = False,
+        project_name: Optional[str] = None,
+        disable_samples: bool = False,
+        filter: SelectorFilterSchema = SelectorFilterSchema(),
+        env: Optional[str] = None,
+        warehouse_type: Optional[str] = None,
+        threads: int = 4,
```
dtaniwaki: Addressed in 7b2df2b — made `threads` a required parameter on `_get_report_data_parallel` so the misleading `=4` default is gone and callers must come through the public router.

elementary/monitor/api/report/report.py
```diff
@@ -136,98 +191,278 @@ def get_report_data(
...
+                f_singular_tests = pool.submit(_new_tests_api().get_singular_tests)
+                f_models_runs = pool.submit(
+                    _new_models_api().get_models_runs,
+                    days_back,
+                    exclude_elementary_models,
+                )
+                f_coverages = pool.submit(_new_models_api().get_test_coverages)
+                f_tests = pool.submit(_new_tests_api().get_tests)
+                f_test_invocation = pool.submit(
+                    _new_invocations_api().get_test_invocation_from_filter, filter
```
dtaniwaki: Addressed in 7b2df2b — switched `pool.submit` calls to `functools.partial(fn, kwarg=...)` so keyword binding survives future signature changes. Same change applied to `f_test_results` / `f_exposures` / `f_test_invocation` further down.

elementary/monitor/api/report/report.py
```diff
@@ -136,98 +191,278 @@ def get_report_data(
...
+                models_runs=models_runs,
+                coverages=coverages,
+                tests=tests,
+                test_invocation=test_invocation,
+                test_results=test_results,
+                source_freshness_results=source_freshness_results,
+                test_runs=test_runs,
+                source_freshness_runs=source_freshness_runs,
+                lineage=lineage,
+                filters_api=FiltersAPI(dbt_runner=parallel_runner),
```
dtaniwaki: Addressed in 7b2df2b — both `LineageAPI` (Phase 3) and `FiltersAPI` (Phase 4) now use `self.dbt_runner` since they are invoked sequentially after the thread pool exits, avoiding unnecessary subprocess spawn overhead.

elementary/monitor/api/report/report.py
```diff
@@ -136,98 +191,278 @@ def get_report_data(
...
+                models_latest_invocation=models_latest_invocation,
+                invocations_data=invocations_data,
             )
-            return report_data, None
         except Exception as error:
             return ReportDataSchema(), error
```
dtaniwaki, line 386: Addressed in 7b2df2b — both the parallel and sequential `except Exception` blocks now call `logger.exception(...)` with a distinct message before returning, so the traceback is preserved for diagnosis.

elementary/monitor/api/report/report.py
```diff
@@ -136,98 +191,278 @@ def get_report_data(
...
+        test_invocation,
+        test_results,
+        source_freshness_results,
+        test_runs,
+        source_freshness_runs,
+        lineage,
+        filters_api,
+        models_latest_invocation,
+        invocations_data,
+    ) -> Tuple[ReportDataSchema, Optional[Exception]]:
```
dtaniwaki, line 412: Addressed in 7b2df2b — `_assemble_report_data` now has full type annotations for all 22 parameters: `Dict[str, NormalizedSeedSchema]`, `ModelRunsWithTotalsSchema`, `DbtInvocationSchema`, `LineageSchema`, etc. Added matching imports.

elementary/monitor/cli.py
```diff
@@ -660,6 +669,13 @@ def report(
     default=None,
     help="Include additional information at the test results summary message.\nCurrently only --include descriptions is supported.",
 )
+@click.option(
+    "--threads",
+    type=click.IntRange(min=1),
+    default=1,
+    help="Number of threads for fetching report data in parallel. "
+    "When set to >1, independent dbt operations run concurrently using subprocess-based runners.",
+)
```
dtaniwaki, line 678: Not addressing for now — only two declarations exist today (`report` and `send-report`). Lifting into `common_options(...)` with conditional gating felt like premature abstraction for a 2-call duplication. Happy to revisit if a third command adopts `--threads`.

tests/unit/monitor/api/report/test_report_parallel.py
```diff
@@ -0,0 +1,125 @@
...
+    runner.target = "dev"
+    runner.raise_on_failure = True
```
dtaniwaki, line 14: Not addressing — `_create_subprocess_runner()` intentionally forwards the parent runner's `raise_on_failure` rather than forcing `True`, so the user's configured behavior carries over to parallel mode. Setting the fixture to `False` while asserting `True` would test a behavior the code does not (and should not) implement. The current test correctly verifies the forwarding contract.

tests/unit/monitor/api/report/test_report_parallel.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+
+    def test_deps_not_run(self, mock_dbt_runner):
+        api = ReportAPI(mock_dbt_runner)
+        with patch(
+            "elementary.monitor.api.report.report.SubprocessDbtRunner"
+        ) as mock_cls:
+            api._create_subprocess_runner()
+            call_kwargs = mock_cls.call_args[1]
+            assert call_kwargs["run_deps_if_needed"] is False
```
dtaniwaki: Addressed in 7b2df2b — `test_deps_not_run` removed. `run_deps_if_needed=False` is already covered by `test_creates_runner_with_correct_config` via `assert_called_once_with(...)`.

PR #2206, opened 2026-04-25
security: remove vulnerable update_pylon_issue.yml workflow

haritamar:
## Summary

- Deletes `.github/workflows/update_pylon_issue.yml`, which contained a GitHub Actions script-injection vulnerability.
- The step `echo \"Comment Body: \${{ github.event.comment.body }}\"` interpolated the comment body directly into a shell `run:` block, letting any GitHub user execute arbitrary code on the runner with the workflow's `GITHUB_TOKEN` (and other secrets) by posting a crafted comment.
- This vector was exploited on **2026-04-24 22:10 UTC** by a throwaway account, which used the runner's token to create spoofed PRs (#2202, #2203, #2204) and trigger the release workflow — publishing a malicious `elementary-data 0.23.3` to PyPI.

## Why a full delete (not a patch)

The workflow's only function was flipping a Pylon ticket state to `waiting_on_you` on every issue/PR comment. Removing it stops the bleeding now; if the Pylon integration is still wanted, it should be reintroduced separately with:

- env-var indirection (`env: COMMENT_BODY: \${{ github.event.comment.body }}` + `\"\$COMMENT_BODY\"` in the script), and
- an `author_association` gate (e.g. only `OWNER`/`MEMBER`/`COLLABORATOR`) so untrusted commenters can't trigger it.

## Follow-ups not in this PR

- Yank/delete `elementary-data 0.23.3` on PyPI.
- Inspect/remove the Docker image pushed by run [24914446630](https://github.com/elementary-data/elementary/actions/runs/24914446630).
- Rotate any secrets exposed to that run (PyPI publish token, Docker registry creds, `PYLON_API_KEY`).
- Delete leftover branches `api-automated-upd-1777068796`, `api-automated-upd-1777069005`, `api-spoof-upd-1777069022` and close PRs #2202, #2203, #2204.
- Audit other workflows for the same `\${{ github.event.* }}` → `run:` pattern.

## Test plan

- [ ] Confirm no other workflow depends on this file.
- [ ] Confirm Pylon ticket state automation loss is acceptable until a hardened replacement lands.

## Summary by CodeRabbit

* **Chores**
  * Removed automated workflow that previously synchronized pull request activity with an external issue tracking system.


magf:
Thank you for catching the vulnerability and acting quickly. That's much appreciated!

I have a small concern about the decision to fully delete the workflow rather than fixing it in place. Since the fix is straightforward (moving the comment body to `env:` + using `"$COMMENT_BODY"`), it would keep the Pylon automation alive without the security risk. Deleting the file also requires reviewing potential breakage elsewhere, and it creates a follow-up task to re‑implement the same feature later.

Not a big deal either way. Just wanted to share a perspective. Thanks again for the quick response on this!

PR #2235, opened 2026-05-16
fix: skip alerts with unknown Status instead of crashing

Darkandz:
## Summary
`apply_filters_schema_on_alert` constructs `Status(alert.data.status)` unconditionally at `elementary/monitor/api/alerts/alert_filters.py:53`. When a row in `get_pending_alerts` carries `status='none'` (e.g., a model that was selected but not actually executed in the dbt run), the whole `edr monitor` invocation crashes with `ValueError: 'none' is not a valid Status`. One bad row poisons the entire alert batch — no alert is delivered.

## Fix
Wrap the `Status(...)` call in `_safe_parse_status`. On unknown values it logs a warning and the alert is filtered out (returns `False`). This matches the behavior the alert would have had anyway under the default status filter (`FAIL/ERROR/RUNTIME_ERROR/WARN`) had `'none'` been a recognized value — except now the rest of the batch is unaffected.

## Reproduction
A real `get_pending_alerts` row from a ClickHouse-backed warehouse with `status='none'` (and `model_unique_id`, `materialization`, `full_refresh` all `null`) triggers the original stack trace at `alert_filters.py:53` → `Status('none')`. After the fix the row is dropped with a logged warning; other alerts in the same batch are sent normally.

## Test plan
- [x] Added `test_filter_alerts_skips_unknown_status` covering both the default status filter and an empty status filter.
- [x] \`pytest tests/unit/monitor/api/alerts/test_alert_filters.py\` — 9 passed.
- [x] \`pre-commit\` hooks (black, isort, flake8, mypy) clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Alerts with unrecognized, missing, or malformed status values are now gracefully skipped during filtering instead of causing errors, improving reliability of alert processing.

* **Tests**
  * Added unit tests verifying invalid alert statuses are dropped while valid alerts continue to be processed under default and empty-status filter conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elementary-data/elementary/pull/2235?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


arbiv (approved):

Darkandz:
@arbiv Hello, i saw that new version of Elementary was realised, but my PR is not included. 

What i need to do ?

elazarlachkar:
> @arbiv Hello, i saw that new version of Elementary was realised, but my PR is not included.
> 
> What i need to do ?

Sorry for the miss @Darkandz. Your change is now released as part of v0.24.1.

PR #2248, opened 2026-06-01
Support custom paths for internal dbt project via env vars

michael-myaskovsky:
## Summary by CodeRabbit

* **New Features**
  * CLI now respects EDR_INTERNAL_TARGET_PATH and DBT_PACKAGES_FOLDER to control where its internal dbt project writes compiled artifacts and installs packages; cleanup targets the configured path.

* **Documentation**
  * Added docs with examples for setting these environment variables and a note clarifying they are separate from the CLI's external --target-path option.

### DIFF
--- a/docs/oss/cli-commands.mdx
+++ b/docs/oss/cli-commands.mdx
@@ -211,3 +211,20 @@ edr monitor --suppression-interval [int]
 ```shell
 edr monitor --override-dbt-project-config --slack-token <your_slack_token> --slack-channel-name <slack_channel_to_post_at> --suppression-interval 0
 ```
+
+## Internal dbt project directories
+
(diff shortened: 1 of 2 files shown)


MikaKerman:
### `DBT_TARGET_PATH` conflicts with `upload_source_freshness`

`DBT_TARGET_PATH` is already used in `upload_source_freshness.py` with a different meaning — it overrides where `edr` looks for the **user's** `sources.json`:

https://github.com/elementary-data/elementary/blob/b82b7d49955110745b09e38a2c31b1f503bbb067/elementary/operations/upload_source_freshness.py#L77-L88

Setting `DBT_TARGET_PATH=/var/tmp/edr_target` to fix internal project permissions will also cause `edr upload-source-freshness` to look for `sources.json` there instead of in the user's project target directory — silently breaking that command.

Consider renaming to something like `EDR_INTERNAL_TARGET_PATH`, or at minimum document the side effect.

### PR title

"fix file permissions issue" doesn't describe what changed — something like "Support custom paths for internal dbt project via env vars" would be clearer.

MikaKerman (approved):

PR #2292, opened 2026-07-08
Pin PyAthena <3.35 for Athena CI

arbiv:
## Summary
- Pin `pyathena>=2.25,<3.35` in `pyproject.toml` as part of the `athena` extra (and `all`)
- Add shared `tests/ci/athena-constraints.txt` for workflows that install dbt without Elementary
- Use pip constraints in `test-warehouse.yml` and `cleanup-stale-schemas.yml` instead of ad-hoc `pip install pyathena...` steps

## Context
PyAthena 3.35.0 refactored internal `_execute()` to use `ExecuteOptions`, breaking dbt-athena 1.10.2 (`work_group` kwarg error). This is an interim pin until dbt-athena ships the boto3 connection manager ([dbt-labs/dbt-adapters#1637](https://github.com/dbt-labs/dbt-adapters/pull/1637)).

## Test plan
- [ ] Athena warehouse CI job passes on this PR
- [ ] Cherry-pick to `release/v0.25.1` to unblock the release run

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI reliability for Athena-based jobs by using an Athena-compatible Python package version for the required driver.
  * Reduced workflow failures during schema cleanup and warehouse testing by applying Athena-specific installation constraints.
  * Updated Athena dependency handling to include the required package version range and ensure “all”/“athena” extras install correctly.

### DIFF
--- a/.github/workflows/test-warehouse.yml
+++ b/.github/workflows/test-warehouse.yml
@@ -247,7 +247,12 @@ jobs:
             DBT_ADAPTER_VERSION_SPEC="~=$DBT_VERSION"
           fi
 
-          pip install "$DBT_CORE_SPEC" "dbt-${DBT_ADAPTER}${DBT_ADAPTER_EXTRA}${DBT_ADAPTER_VERSION_SPEC}"
+          CONSTRAINTS_ARGS=()
+          if [ "$WAREHOUSE_TYPE" = "athena" ]; then
+            CONSTRAINTS_ARGS=(-c tests/ci/athena-constraints.txt)
+          fi
+
+          pip install "${CONSTRAINTS_ARGS[@]}" "$DBT_CORE_SPEC" "dbt-${DBT_ADAPTER}${DBT_ADAPTER_EXTRA}${DBT_ADAPTER_VERSION_SPEC}"
 
       # dbt-vertica pins dbt-core~=1.8 which lacks the 'arguments' attribute
       # used by newer dbt-core.  Install dbt-vertica without deps first, then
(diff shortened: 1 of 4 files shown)


.github/workflows/test-warehouse.yml
```diff
@@ -247,7 +247,12 @@ jobs:
             DBT_ADAPTER_VERSION_SPEC="~=$DBT_VERSION"
           fi
 
-          pip install "$DBT_CORE_SPEC" "dbt-${DBT_ADAPTER}${DBT_ADAPTER_EXTRA}${DBT_ADAPTER_VERSION_SPEC}"
+          CONSTRAINTS_ARGS=()
+          if [ "$WAREHOUSE_TYPE" = "athena" ]; then
+            CONSTRAINTS_ARGS=(-c tests/ci/athena-constraints.txt)
```
elazarlachkar: Won't the constraint be installed in the `Install Elementary`, as it's in the `pyproject.toml`? Why do we need to install it explicitly here?

arbiv: Good catch — you're right for `test-warehouse.yml`. The `pyproject.toml` pin via `pip install ".[athena]"` in the next step is sufficient, so I removed the constraints from Install dbt.

We still need `tests/ci/athena-constraints.txt` in `cleanup-stale-schemas.yml`, which installs `dbt-athena-community` only and never installs Elementary, so the `pyproject.toml` extra doesn't apply there.

elazarlachkar (approved):

Issue #2283, opened 2026-06-26
--report-url deep links for test alerts open the full test list instead of the specific test (cold-load redirect on #/report/test-runs/<id>)


niccoloalexander:
### Summary
Deep links to `#/report/test-runs/<test_unique_id>/` don't resolve on a cold page load — they redirect to `#/report/test-runs` (the full Test Performance list). This is the exact link `edr monitor --report-url` generates for **test alerts**, so every Slack alert's report link opens the full test list instead of the failing test.

This is the same class of cold-load redirect-guard bug as #2182, but for **test-runs** (that issue is scoped to model-runs and notes test-runs as unaffected — that's no longer the case).

### Environment
- edr (elementary-data): reproduced on **0.23.1, 0.23.2, 0.23.4, 0.24.0, 0.24.1, 0.25.0**
- OSS self-hosted report (static HTML, no Cloud)

### Steps to reproduce
1. Generate a report and send alerts with `edr monitor --report-url <host>`.
2. In a **fresh tab**, open a test alert's report link: `<host>/#/report/test-runs/<test_unique_id>/`.
3. Observe it lands on `#/report/test-runs` (full list); the specific test is never selected.
4. Pasting the same URL into an **already-open** report tab works — same warm/cold distinction as #2182.

### Expected vs actual
- **Expected:** the link opens the specific test's run history.
- **Actual:** redirects to the full Test Performance list.

### Evidence across versions
I injected a real report's `window.elementaryData` into each version's bundled `report/index.html` and cold-loaded the deep link in headless Chrome. All six versions (0.23.1 → 0.25.0) land on the unfiltered list. The data renders correctly in each, so it's the routing/redirect guard, not a data problem.

### Likely cause / fix
Same mechanism as #2182: the detail route's redirect guard fires while data is still in its initial empty/fetching state, then navigates to the parent list. Likely fixes: guard the redirect `useEffect` with the `isFetching` flag, or fall back to embedded report data instead of an empty array during fetch.

(Noting the report frontend that generates the artifact isn't public, so external contributors can't patch it directly.)

niccoloalexander:
Dug into this a bit further while working around it. Summary of the root cause and the layers involved, in case it helps whoever picks it up:

[...]

PR #2296, opened 2026-07-10
fix(alerts): route test alert deep links to the cold-load-safe test-results route (#2283)

niccoloalexander:
Fixes the broken cold-load behaviour reported in #2283.

## Problem

Test alert deep links (`edr monitor --report-url ...`) point at `#/report/test-runs/<id>`. Opening one from a Slack alert is always a **cold page load** (fresh tab), and on a cold load the detail route's redirect guard fires while report data is still fetching (empty array) and navigates to the parent Test Performance list — so the alert never opens the specific failing test. Pasting the same URL into an already-open report tab works, which is why it only reproduces from alerts. Affects 0.23.1 → current `master`.

## Change

Reroute test alert deep links to `#/report/test-results/<id>`, which does not hit the guard and cold-loads to the correct test (it also surfaces that test's run history). One-line change in `report_link.py`.

## Caveat (maintainers' call)

This is a **link-layer fix for the symptom**. The underlying bug is the redirect guard in the report UI; fixing it directly (e.g. gating the redirect `useEffect` on `isFetching`, or falling back to embedded report data during fetch) would let `test-runs/<id>` keep working, and might be your preferred fix. I don't have access to the report UI source (this repo ships the compiled `index.html`), so this is the fix I can offer from the public repo. Happy to adjust — e.g. also retext the "View test runs" button so its label matches the results destination — if you'd like.

Verified: with this change, `edr monitor` regenerates the working `test-results/<id>` link and the alert opens the specific failing test on a cold load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated report-history URL generation to use the correct `test-results` path for test runs.
  * Improved link behavior for direct page loads of individual test runs to avoid cold-load issues.

### DIFF
--- a/elementary/monitor/data_monitoring/alerts/integrations/utils/report_link.py
+++ b/elementary/monitor/data_monitoring/alerts/integrations/utils/report_link.py
@@ -15,7 +15,8 @@ class ReportLinkData(BaseModel):
 
 
 class ReportPath(Enum):
-    TEST_RUNS = "test-runs"
+    # test-runs/<id> cold-loads to the parent list, not the test (see #2283)
+    TEST_RUNS = "test-results"
     MODEL_RUNS = "model-runs"
 
 
--- a/tests/unit/alerts/test_report_link.py
+++ b/tests/unit/alerts/test_report_link.py
@@ -1,0 +1,44 @@
+from elementary.monitor.data_monitoring.alerts.integrations.utils.report_link import (
+    ReportPath,
+    get_model_runs_link,
+    get_model_test_runs_link,
+    get_test_runs_link,
+)
+
+REPORT_URL = "https://elementary.example.com/#"
+TEST_UNIQUE_ID = "test.my_project.my_test.abc123.row_count"
+MODEL_UNIQUE_ID = "model.my_project.my_model"
+
+
+def test_get_test_runs_link_routes_to_test_results():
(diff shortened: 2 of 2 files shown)


niccoloalexander:
Added tests in `tests/unit/alerts/test_report_link.py` covering the link builders: test alert deep links now resolve to `/report/test-results/<id>/` (and never `/report/test-runs/`), the button label is unchanged, model-run links are untouched, and the `None` cases.

No docs changes: this is a bug fix to the destination of an existing deep link (not a new user-facing feature or config), and the docs don't reference the report route paths. Happy to add a note somewhere if you'd prefer.

niccoloalexander:
Thanks @coderabbitai. On the `TEST_RUNS` → `TEST_RESULTS` rename nitpick — I've left the member name as-is on purpose:

This is an **interim** reroute. The intended end-state is fixing the report-UI redirect guard (#2283), after which `TEST_RUNS` should map back to `"test-runs"`. Keeping the member name (plus the inline comment) signals "this is the test-runs link, temporarily pointed at the cold-load-safe route" rather than a permanent semantic change, and keeps the diff to one line instead of rippling into `slack.py` and the `get_test_runs_link` / `get_model_test_runs_link` helpers.

Happy to do the rename if a maintainer would rather treat this as the permanent behaviour.

haritamar:
Closing, will fix in the report separately.

PR #2303, opened 2026-07-20
fix: normalize non-standard fractional-second precision when parsing timestamps (#2221)

### LINKED ISSUE #2221: Different timestamp precision on macOS vs Linux causes BigQuery timestamp failure
Hi everyone,

I came across an interesting issue and wanted to check if anyone has seen something similar.

I am using the following versions:

```
elementary-data==0.20.1 cli and python packages
dbt-bigquery==1.10.2
dbt-fusion==2.0.0-preview.173
dbt-core==1.11.0b3

```

## Issue

When running `dbt test` locally on **macOS**, the `execute_completed_at` column is generated with this timestamp format:

```text
2026-04-27T14:05:11.371056Z
```

However, when running the **same test**, with the **same package versions**, inside **Docker (Linux)**, the precision changes to:

```text
2026-04-27T14:34:33.609083964Z
```

The second value includes nanosecond precision, and BigQuery rejects it as an invalid timestamp, which later causes an error in the `edr report`.

## Question

Does anyone know why the timestamp precision differs between macOS and Linux even with identical package versions?

It seems the environment/OS may be affecting how the timestamp is generated, but I am not sure which component is responsible.

Any ideas would be appreciated.

zerafachris:
Fixes #2221

## Root cause

`datetime.fromisoformat()` only reliably accepts a fractional-seconds component of exactly 0, 3, or 6 digits on Python versions before 3.11 (the only widths `datetime.isoformat()` itself ever produces); any other width — including 9-digit/nanosecond-precision timestamps like the one reported in the issue — raises `ValueError` there, while Python 3.11+ tolerates (and truncates) any digit count. Since `elementary` supports `python >= 3.10`, `convert_partial_iso_format_to_full_iso_format()` in `elementary/utils/time.py` (used to parse `generated_at`/`detected_at`/`time_utc` timestamps coming from dbt/warehouse metadata) would hit that `ValueError`, get caught, log an exception, and silently fall back to returning the **raw, unnormalized** timestamp string. That's a Python-version-dependent code path, not truly an OS-dependent one, but it reproduces the "works on macOS, breaks in the Linux/Docker container" symptom whenever the two environments happen to run different Python builds — one on each side of the 3.11 boundary — and the retained over-precise value is what later trips up downstream consumers (e.g. report generation / BigQuery).

## What changed and why

Added `_normalize_fractional_seconds()` to `elementary/utils/time.py`, which coerces the fractional-seconds component of the input string to exactly 6 digits (truncating extra precision, right-padding a shorter one) *before* calling `datetime.fromisoformat()`. This is called from `convert_partial_iso_format_to_full_iso_format()` right after the existing timezone-offset normalization step. The fix makes parsing behavior explicit and consistent across all supported Python versions instead of depending on interpreter-specific `fromisoformat` leniency. No other behavior changes — the function still always drops microseconds in its output (as it already did), so all pre-existing test cases are unaffected.

## Test evidence

Added `tests/unit/utils/test_time.py::test_convert_partial_iso_format_to_full_iso_format_handles_any_fractional_precision`, parametrized over nanosecond-precision (`Z` suffix and full offset), single-digit, and five-digit fractional-second inputs. Since the bug only manifests on Python < 3.11's stricter `fromisoformat`, the test forces that exact legacy behavior deterministically (a `datetime` subclass overriding `fromisoformat` to replicate CPython's pre-3.11 restriction) so it fails reliably regardless of which Python version runs the suite.

- Confirmed the new tests fail before the fix (all 4 cases raise/mis-handle the `ValueError` path, e.g. `AssertionError: assert '2026-04-27T14:34:33.609083964Z' == '2026-04-27T14:34:33+00:00'`).
- After the fix: `python -m pytest tests/unit/utils/test_time.py -vv` → **18 passed**.
- Full suite: `python -m pytest tests/unit -q` → **455 passed**.
- Also manually verified against a real Python 3.9 interpreter (which has the actual restrictive `fromisoformat`, not just the mocked version) that the normalization regex fixes real parsing of 9-digit and 1-digit fractional timestamps.
- Lint/type-check on touched files: `black --check`, `isort --profile black --check`, `flake8`, and `mypy --no-error-summary` all clean.

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of ISO-8601 timestamps with fractional seconds of varying lengths.
  * Prevented parsing failures for timestamps with nanosecond, millisecond, or other fractional precision.
  * Ensured consistent timestamp conversion across supported Python versions.


zerafachris:
Tests were added in this PR (4 parametrized cases in `tests/unit/utils/test_time.py` covering nanosecond-precision, non-standard fractional-second widths, both pre- and post-3.11 `fromisoformat` behavior). This is an internal timestamp-parsing bug fix with no user-facing behavior, config, or API change, so there's nothing in `docs/` that needs updating for it.

haritamar (approved):

PR #2307, opened 2026-07-23
fix: use testName/testSubType filters instead of hidden testCategory column

### LINKED ISSUE #2294: Dashboard click-through filter uses hidden "testCategory" column instead of "testSubType", returning empty Test Results
Elementary version: 0.25.0 / dbt package 0.25.1

Adapter: Snowflake

Description

Clicking a test category tile on the Dashboard (e.g. "Anomalies", "Volume", "Schema changes") navigates to the Test Results tab with a filter applied, but the tab appears empty.

Root cause (identified in the bundled JS)

The click-through handler (Y0n) generates the filter as:

testCategory is defined as a column in the Test Results table but has hideColumn: true and is never populated from the row data — the data mapping only includes testType and testSubType. As a result the filter matches no rows.

The filter chip displayed to the user reads "Test category is anomalies", but the correct filterable column is "Test type" (id: "testSubType").

Steps to reproduce

Run edr report with at least one Elementary test result present
Open the HTML report in a browser
On the Dashboard tab, click any test category tile (e.g. the Anomalies count)
Observe: Test Results tab opens with "Test category is anomalies" filter applied and zero rows shown
Expected behaviour

Test Results tab opens filtered to the matching test type, showing the relevant rows.

Workaround

Manually delete the "Test category" filter chip and add a "Test type" filter with the equivalent value.

sravankumarkunadi:
## Description

Fixes #2294

Clicking a test category tile on the Dashboard (e.g. "Anomalies", "Volume") navigates to the Test Results tab with a `testCategory` filter applied. However, `testCategory` is a hidden column that is never populated from row data, so the filter matches no rows and the tab appears empty.

This removes the `testCategory` shortcut from `Y0n` and falls back to the existing `testName`/`testSubType` filter logic, which correctly matches the data in the Test Results table.

## Note on the changed file

The only changed file is `elementary/monitor/data_monitoring/report/index.html`, which is the built report bundle, so the diff is a one-token change in minified code. That means the next UI build will regenerate it and drop this fix unless the same change also lands in the UI source. Happy to open that PR too if you point me at the right repo — otherwise treat this as the stopgap for the shipped report.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated test filtering to rely on test name and optional subtype criteria.
  * Removed automatic filtering by test category to improve filter accuracy.


sravankumarkunadi:
This is a patch to the bundled  index.html no source-level tests or docs apply here since the UI is built separately. Happy to add anything if needed though!

sravankumarkunadi:
@arbiv can I please get a review on this one?

haritamar:
Closing this one, we'll add an updated bundled report instead

PR #2308, opened 2026-07-27
fix: use explicit utf-8 encoding and LF newlines for file I/O

sravankumarkunadi:
## What

Pass `encoding="utf-8"` explicitly to every `Path.read_text()` / `Path.write_text()` call in `scripts/generate_report_from_json.py` and `tests/unit/messages/utils.py`, and `newline="\n"` to the fixture writes.

## Why

`read_text()`/`write_text()` fall back to `locale.getpreferredencoding()` and platform newline translation when no encoding is given. That's UTF-8/LF on Linux and macOS, but cp1252/CRLF on Windows.

**Before**
- `scripts/generate_report_from_json.py` raises `UnicodeDecodeError` on Windows when the report template (`elementary/monitor/data_monitoring/report/index.html`) or the JSON output contains any non-cp1252 character.
- `tests/unit/messages/utils.py` run with `OVERRIDE=true` on Windows rewrites the expected JSON/text fixtures with CRLF line endings, producing a diff on every fixture file that touches the repo.

**After**
- Both files are read and written as UTF-8 on every platform.
- Regenerated fixtures keep LF line endings regardless of the platform they were generated on.

## Notes

No behavior change on Linux/macOS — this only makes the already-assumed defaults explicit. `Path.write_text(newline=...)` requires Python 3.10, which matches this project's `python = ">=3.10,<3.14"`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report generation consistency across platforms by standardizing UTF-8 file encoding.
  * Normalized line endings in generated reports and expected test files for more reliable output.

### DIFF
--- a/scripts/generate_report_from_json.py
+++ b/scripts/generate_report_from_json.py
@@ -3,7 +3,7 @@ import sys
 
 
 def gen_report(json_output_path: str):
-    json_output_data = pathlib.Path(json_output_path).read_text()
+    json_output_data = pathlib.Path(json_output_path).read_text(encoding="utf-8")
     report_template_path = (
         pathlib.Path(__file__).parent.parent
         / "elementary"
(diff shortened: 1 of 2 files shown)


sravankumarkunadi:
Closing this — opened in error, unrelated to the work I'm actually doing on this repo. Sorry for the noise.

PR #2323, opened 2026-08-12
CORE-1273: migrate from Poetry to uv

haritamar:
## Summary

Replaces Poetry with uv as the packaging/dev tool. `pyproject.toml` becomes standard PEP 621 metadata built by `uv_build`, dev deps move from `dev-requirements.txt` into a uv dependency group, and `uv.lock` is committed (uv's default) so dev/CI resolution is reproducible and the full graph is auditable — `run-precommit` uses `uv sync --locked` and a `uv-lock` pre-commit hook keeps it in sync.

Because the package lives at the repo root rather than in `src/`, the build backend needs:

```toml
[build-system]
requires = ["uv_build>=0.9.0,<0.15.0"]
build-backend = "uv_build"

[tool.uv.build-backend]
module-root = ""
module-name = "elementary"
```

Extras (`snowflake`, `bigquery`, ..., `all`) and the `edr` script are unchanged in behavior; the `all` extra is now expressed as a self-referential `elementary-data[...]` list, which is the PEP 621 equivalent of Poetry's extras aggregation.

**`constraint-dependencies`:** `urllib3`, `idna`, `pyasn1` and `cryptography` are transitive deps that were only listed to dodge CVEs, so they move out of `[project.dependencies]` into constraints — downstream consumers scan their own locks, while everything we ship and test still gets the floors (`uv.lock`, CI, the image):

```toml
[tool.uv]
constraint-dependencies = ["urllib3>=2.7.0", "idna>=3.15", "pyasn1>=0.6.4", "cryptography>=50.0.0"]
```

Constraints are read from the pyproject in uv's *working directory*, not from the install target, so the Dockerfile installs with `--directory /app ".[all]"` instead of `"/app[all]"` — otherwise the image would resolve unconstrained. Verified in the built image: `urllib3 2.7.0`, `idna 3.18`, `pyasn1 0.6.4`, `cryptography 50.0.0`.

CI/dev surface:
- `run-precommit`: `setup-python` + pip → `astral-sh/setup-uv` + `uv sync --locked` + `uv run pre-commit`.
- `test-warehouse` / `test-github-action`: `pip install` → `uv pip install` (with `UV_SYSTEM_PYTHON=1`), dev deps via `uv pip install --group dev`; the dbt install retries 3x because `dbt-core-experimental-parser`'s build backend flakily downloads its wheel from GitHub releases.
- `release`: `pip install build` + `python -m build` → `uv build --sdist --wheel`.
- `Dockerfile`: uv `0.10.11` → `0.12.3`.
- `dependency-review-config.yml`: committing the lock makes the action review the whole resolved graph, so `metricflow` (Apache-2.0; the AGPL/BSL strings are its README's historical-licensing note), `numpy` (`BSD-3-Clause AND 0BSD AND MIT AND Zlib AND CC0-1.0`, which the action doesn't decompose) and `rapidfuzz` (MIT; GPL headers come from vendored FastFlow examples not built into the wheel) join the existing `lxml`/`typing-extensions` exemptions. Package-level, so GPL-only additions are still blocked.

Verified locally: `uv sync --all-extras`, 451 unit tests pass, all pre-commit hooks pass, `uv build` artifacts contain the `elementary` package plus the bundled dbt project assets and no longer publish the transitive floors, wheel installs and `edr --help` works, and `docker build` + container run succeed.

Link to Devin session: https://app.devin.ai/sessions/c214461411ba46d9ab639b8ea12e69b5
Requested by: @haritamar

## Summary by CodeRabbit

* **Chores**
  * Migrated project and development dependency management to uv.
  * Updated build, test, pre-commit, and release workflows to use a consistent, pinned uv setup with caching.
  * Updated the container build environment to uv 0.12.3.
  * Consolidated development dependencies into project configuration and removed the separate development requirements file.
  * Added automatic lockfile validation to pre-commit checks.
  * Standardized project packaging and optional adapter dependency configuration.

* **Documentation**
  * Updated contributor setup and command instructions for the new uv-based workflow.


.github/workflows/release.yml
```diff
@@ -11,6 +11,7 @@ on:
       - v[0-9]+.[0-9]+.[0-9]+
 
 env:
+  UV_VERSION: "0.12.1"
```
haritamar: I think the latest is 0.12.3, perhaps we should use that?

.github/workflows/release.yml
```diff
@@ -106,16 +107,13 @@ jobs:
         with:
           ref: ${{ needs.validate-release-ref.outputs.release-ref }}
 
-      - name: Setup Python
-        uses: actions/setup-python@v6
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
         with:
-          python-version: "3.10"
```
haritamar, line 112: Should we set UV_PYTHON in this workflow, like I see you did in other workflows?

.pre-commit-config.yaml
```diff
@@ -21,6 +21,11 @@ repos:
       - id: prettier
         exclude: \.html$|^docs/|^tests/unit/messages/formats/markdown/fixtures/
 
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: 0.12.1
+    hooks:
+      - id: uv-lock
```
haritamar: Should we actually add uv.lock to git, given that it's a published python package?
I'm actually not sure what's the best practice.

pyproject.toml
```diff
@@ -1,94 +1,119 @@
...
 description = "Data monitoring and lineage"
-authors = ["Elementary"]
+authors = [{ name = "Elementary" }]
 keywords = ["data", "lineage", "data lineage", "data warehouse", "DWH", "observability", "data monitoring", "data observability", "Snowflake", "BigQuery", "Redshift", "data reliability", "analytics engineering"]
 license = "Apache-2.0"
+license-files = ["LICENSE"]
 readme = "README.md"
-homepage = "https://www.elementary-data.com"
-repository = "https://github.com/elementary-data/elementary"
-documentation = "https://docs.elementary-data.com"
```
haritamar, line 11: Should we keep these metadata fields? why did you decide to remove them? they're not supported for uv?

.pre-commit-config.yaml
haritamar: Let's gitignore and drop the hook for now

PR #2333, opened 2026-08-23
Support dbt 2.0 / Fusion in the edr CLI

haritamar:
## Summary

Adds support for dbt 2.0 / Fusion to the `edr` CLI (CORE-1344). dbt 2.0 is the Fusion engine, installable via pip in two flavors: `dbt-core>=2.0` (Python wheel, no callback API) and `dbt` (binary-only platform wheel, no importable Python module at all). Both are driven via subprocess by a new `Dbt2Runner`.

**Runner auto-detection** — `factory.py` no longer imports `dbt.version` at import time (which crashes with binary-only Fusion). Instead, the flavor is detected lazily from package metadata (new `dbt_installation.py`):

```
DBT_RUNNER_METHOD env var          -> explicit override (unchanged)
dbt-core >= 2.0                    -> Dbt2Runner
dbt-core >= 1.5                    -> APIDbtRunner (unchanged)
dbt-core <  1.5                    -> SubprocessDbtRunner (unchanged)
no dbt-core, fusion binary present -> Dbt2Runner
nothing detected                   -> SubprocessDbtRunner
```

`RunnerMethod.DBT2` (`"dbt2"`) is added; `RunnerMethod.FUSION` (`"fusion"`) is kept as a backward-compatible alias (dbt-data-reliability CI passes `--runner-method fusion`), and `dbt_fusion_runner.py` re-exports `Dbt2Runner` as `DbtFusionRunner`.

**`Dbt2Runner`** replaces `DbtFusionRunner`: resolves the binary via `DBT_FUSION_PATH` env var → `dbt` on PATH (only when it can't be a dbt-core 1.x entrypoint) → `~/.local/bin/dbt`, and no longer skips `dbt deps` (Fusion supports it).

**Fusion compatibility fixes**:
- Fusion rejects `target-path` in `dbt_project.yml`, so it was removed from the internal project; `EDR_INTERNAL_TARGET_PATH` is now translated by `CommandLineDbtRunner` into the standard `DBT_TARGET_PATH` env var when running the internal project.
- The e2e dbt project was migrated to dbt 2.0-compatible syntax with `dbt-autofix` (`meta`/`tags`/test args moved under `config`/`arguments` in `schema.yml`), plus a generic test definition for the previously-implicit `uniques` test. `+root_path` (a dbt-dremio 1.x config that Fusion rejects) is kept for the Dremio target and stripped by CI for dbt 2.x targets.
- `pyproject.toml`: `dbt-core` widened to `>=1.8,<3.0.0` (2.x requires Python ≥3.11; adapter extras stay 1.x-only since 2.0 has adapters built in).

**CI** — `test-warehouse.yml` accepts dbt-version values:
- `fusion`: installs the binary wheel, pinned to `dbt==2.0.0rc212` (stable `pip install dbt` currently resolves to the unrelated dbt Cloud CLI);
- `2.x`: installs the latest dbt-core 2.x at run time via `pip install --pre "dbt-core>=2.0.0a0,<3"`, so future betas/RCs/the official release are picked up automatically (an explicit version like `2.0.0b3` still works for manual-dispatch pinning);
- both run on Python 3.11 and skip adapter extras.

`test-all-warehouses.yml` gets a separate informational `test-dbt2` matrix on Fusion-supported warehouses (snowflake, bigquery, databricks). dbt 2.x snowflake jobs run with `threads: 1` (profile + explicit `--threads 1`) to work around the Fusion Snowflake driver's concurrent-connection hang (dbt-labs/dbt-fusion#410); the `2.x` + snowflake cell is excluded because the current dbt-core `2.0.0b2` engine hangs on Snowflake even single-threaded — the fusion target covers Snowflake until a newer dbt-core 2.x is released.

Verified locally against DuckDB with Fusion 2.0.0-preview.212: `parse`, `deps`, `seed`, `run-operation` marker capture, `ls` parsing, and the internal target-path translation all work through `Dbt2Runner`. Unit tests (466) pass; new factory/detection tests added in `test_factory.py`. The full CI matrix (all dbt 1.x warehouses + fusion on snowflake/bigquery/databricks + dbt-core 2.x on bigquery/databricks) is green.

Linear: [CORE-1344](https://linear.app/elementary/issue/CORE-1344/support-dbt-20-fusion-in-the-edr-cli)

Link to Devin session: https://app.devin.ai/sessions/496f2bf70ae14ec1a1bccb245fbc9328
Requested by: @haritamar


.github/workflows/test-warehouse.yml
```diff
@@ -133,11 +134,31 @@ jobs:
...
 
-          if [ -n "$DBT_VERSION" ] && ! [[ "$DBT_VERSION" =~ ^[0-9]+(\.[0-9]+){1,2}([a-zA-Z0-9._+-]+)?$ ]]; then
+          if [ -n "$DBT_VERSION" ] && [ "$DBT_VERSION" != "fusion" ] && [ "$DBT_VERSION" != "2.x" ] && ! [[ "$DBT_VERSION" =~ ^[0-9]+(\.[0-9]+){1,2}([a-zA-Z0-9._+-]+)?$ ]]; then
             echo "Unsupported dbt version: $DBT_VERSION" >&2
             exit 1
           fi
 
+          # dbt 2.x / Fusion has adapters built into the engine and doesn't
+          # support third-party adapters such as dbt-vertica.
+          if [ "$WAREHOUSE_TYPE" = "vertica" ] && { [ "$DBT_VERSION" = "fusion" ] || [[ "$DBT_VERSION" == 2* ]]; }; then
```
haritamar: Why do we need a vertica specific check?

.github/workflows/test-warehouse.yml
```diff
@@ -226,7 +247,8 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.10"
+          # dbt 2.x (the Fusion engine) requires Python >= 3.11.
+          python-version: ${{ (inputs.dbt-version == 'fusion' || startsWith(inputs.dbt-version, '2.')) && '3.11' || '3.10' }}
```
haritamar: Then let's simplify and use 3.11 globally, no?

.github/workflows/test-warehouse.yml
```diff
@@ -292,6 +336,20 @@ jobs:
...
+        run: |
+          # Remove dbt-core (pulled in as a dependency of elementary) so the
+          # environment matches a binary-only Fusion installation, then install
+          # the Fusion binary from the 'dbt' package.
+          # Note: without a version pin, 'pip install dbt' resolves to the
+          # unrelated legacy dbt Cloud CLI package, hence the explicit 2.x spec.
+          pip uninstall -y dbt-core
+          # Pinned for reproducible CI runs (rc205, for example, generates
+          # broken BigQuery temp-table DDL: "Invalid timestamp: '{}'").
+          pip install "dbt==2.0.0rc212"
```
haritamar: I'm a bit bummed by pinning a specific version. Maybe that's not an issue anymore?

tests/profiles/profiles.yml.j2
```diff
@@ -112,7 +112,7 @@ elementary_tests:
       database: {{ snowflake_database | toyaml }}
       warehouse: {{ snowflake_warehouse | toyaml }}
       schema: {{ schema_name }}
-      threads: 4
+      threads: {{ snowflake_threads | default(4) }}
```
haritamar: Do we need it to be 4 anywhere? Any reason not to put "1" hear instead of overriding in the workflow?
(asking as this template is specific for the tests)

.github/workflows/test-all-warehouses.yml
```diff
@@ -132,3 +132,43 @@ jobs:
       CI_SLACK_WEBHOOK: ${{ secrets.CI_SLACK_WEBHOOK }}
       CI_SLACK_TOKEN: ${{ secrets.CI_SLACK_TOKEN }}
       AWS_OIDC_ROLE_ARN: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+
+  # dbt 2.0 (Fusion engine) targets, limited to Fusion-supported warehouses.
+  # 'fusion' installs the standalone binary from the 'dbt' PyPI package;
+  # '2.x' installs the latest 'dbt-core' 2.x Python package (including
+  # pre-releases); an explicit 2.x version pins that package version.
+  # dbt 2.0 is still a prerelease; these jobs are separate from 'test' so they
+  # can be treated as informational (not required checks) while it stabilizes.
```
haritamar: I think this comment doesn't add a lot of information, please remove it

.github/workflows/test-warehouse.yml
```diff
@@ -32,7 +32,7 @@ on:
       dbt-version:
         type: string
         required: false
-        description: dbt's version to test with
+        description: dbt's version to test with ('fusion' for the dbt Fusion binary)
```
haritamar: I don't think this addition is necessary

.github/workflows/test-warehouse.yml
```diff
@@ -80,6 +80,7 @@ env:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 90
```
haritamar, line 83: I think we may have done this regardless on prod, but worth verifying

.github/workflows/test-warehouse.yml
```diff
@@ -133,11 +134,31 @@ jobs:
               ;;
           esac
 
-          if [ -n "$DBT_VERSION" ] && ! [[ "$DBT_VERSION" =~ ^[0-9]+(\.[0-9]+){1,2}([a-zA-Z0-9._+-]+)?$ ]]; then
+          if [ -n "$DBT_VERSION" ] && [ "$DBT_VERSION" != "fusion" ] && [ "$DBT_VERSION" != "2.x" ] && ! [[ "$DBT_VERSION" =~ ^[0-9]+(\.[0-9]+){1,2}([a-zA-Z0-9._+-]+)?$ ]]; then
```
haritamar, line 137: Just making sure - if we want to trigger a run with a specific "2.x" version, we can type the exact version and it will still work?

.github/workflows/test-warehouse.yml
```diff
@@ -133,11 +134,31 @@ jobs:
...
+          # support third-party adapters such as dbt-vertica.
+          if [ "$WAREHOUSE_TYPE" = "vertica" ] && { [ "$DBT_VERSION" = "fusion" ] || [[ "$DBT_VERSION" == 2* ]]; }; then
+            echo "dbt version '$DBT_VERSION' is not supported with the vertica warehouse" >&2
+            exit 1
+          fi
+
+      - name: Remove dbt 1.x-only configs from E2E dbt project
+        if: inputs.dbt-version == 'fusion' || startsWith(inputs.dbt-version, '2.')
+        working-directory: ${{ env.E2E_DBT_PROJECT_DIR }}
+        run: |
```
haritamar, line 158: Is there some elegant way we can support these changes directly in the dbt_project.yml file?

.github/workflows/test-warehouse.yml
```diff
@@ -387,24 +463,29 @@ jobs:
           if [ "$WAREHOUSE_TYPE" = "dremio" ]; then
             EXTRA_ARGS+=(--threads 1)
           fi
+          # The dbt 2.x Snowflake driver hangs with concurrent connections
+          # (dbt-labs/dbt-fusion#410); pass --threads 1 explicitly on top of
+          # the profile-level setting as an extra safeguard.
+          if { [ "$DBT_VERSION" = "fusion" ] || [[ "$DBT_VERSION" == 2* ]]; } &&
+             [ "$WAREHOUSE_TYPE" = "snowflake" ]; then
+            EXTRA_ARGS+=(--threads 1)
```
haritamar: also commented below, asking if we should just set threads=1 in the profiles yml file.
If we only want to do it in fusion, it may be more elegant to also pass the version information to jinja.

.github/workflows/test-warehouse.yml
```diff
@@ -320,10 +378,28 @@ jobs:
 
           echo "SCHEMA_NAME=$SCHEMA_NAME" >> "$GITHUB_ENV"
 
+          # The seeds schema is shared between jobs on the same warehouse, and
+          # the dbt 2.x jobs run in parallel with the dbt 1.x job on the same
+          # warehouses - give each dbt version its own seeds schema to avoid
+          # concurrent-write conflicts (e.g. Delta transaction conflicts).
+          if [ "$DBT_VERSION" = "fusion" ] || [[ "$DBT_VERSION" == 2* ]]; then
+            SAFE_DBT_VERSION=$(echo -n "$DBT_VERSION" | sed 's/[^a-zA-Z0-9]/_/g')
+            echo "TEST_SEEDS_SCHEMA=test_seeds_$SAFE_DBT_VERSION" >> "$GITHUB_ENV"
```
haritamar: I'm just wondering - as I recall a part of the reasoning for even specifying the test_seeds schema is that we don't continuously generate it (e.g. the workflow has generate-data=False by default)

So doesn't it mean this schema should actually stay "test_seeds" like before?
Do we actually seed into "test_seeds"?


I guess if you can share a bit more about the case you saw that'd be great.

dev-requirements.txt
```diff
@@ -1,6 +1,9 @@
 pytest
 pytest-parametrization>=2022.2.1
 pre-commit
+# Used by tests/profiles/generate_profiles.py; with dbt 2.x / Fusion it is no
+# longer pulled in transitively by dbt-core.
```
haritamar: remove comment

.github/workflows/test-warehouse.yml
haritamar: ah interesting.
so maybe the better fix is to actually remove the custom seed directory completely?
so that seeds will land in the same branch schema as models?

tests/e2e_dbt_project/models/schema.yml
```diff
@@ -657,7 +657,7 @@ models:
       tags: ["config_levels"]
 sources:
   - name: training
-    schema: "{{ target.schema if target.type == 'dremio' else env_var('TEST_SEEDS_SCHEMA', 'test_seeds') }}"
+    schema: "{{ target.schema }}"
```
haritamar, line 660: isn't `target.schema` the default? or is this still needed?

elementary/monitor/dbt_project/dbt_project.yml
```diff
@@ -20,7 +20,9 @@ snapshot-paths: ["snapshots"]
 
 packages-install-path: "{{ env_var('DBT_PACKAGES_FOLDER', 'dbt_packages') }}"
 
-target-path: "{{ env_var('EDR_INTERNAL_TARGET_PATH', 'target') }}" # directory which will store compiled SQL files
+# NOTE: 'target-path' is intentionally not set here since dbt 2.0 (Fusion) rejects it.
+# The EDR_INTERNAL_TARGET_PATH env var is instead translated to DBT_TARGET_PATH by the CLI
+# when running this project.
```
haritamar: please remove the comment - it's now a comment on nothing, as the original code is gone.

elementary/clients/dbt/dbt_fusion_runner.py
```diff
@@ -1,14 +1,4 @@
-import os
+# Kept for backward compatibility; use Dbt2Runner instead.
```
GuyEshdat: maybe we should mark as deprecated and log warning?

elementary/clients/dbt/factory.py
```diff
@@ -2,19 +2,22 @@
...
+from elementary.clients.dbt.dbt_installation import (
+    get_dbt_core_version,
+    is_dbt2_binary_available,
+)
 from elementary.clients.dbt.subprocess_dbt_runner import SubprocessDbtRunner
 
-DBT_VERSION = version.Version(dbt_version_string)
-
 
 class RunnerMethod(Enum):
```
GuyEshdat, line 16: maybe also add comments to the other methods (api/subprocess) that they support v1 only?

GuyEshdat (approved):

PR #2341, opened 2026-08-31
fix(alerts): treat naive detected_at datetimes as UTC before timezone conversion

### LINKED ISSUE #2304: Alert "Time" shows the UTC value labeled with the local timezone when the host timezone is not UTC
**Describe the bug**

`AlertModel.__init__` converts `detected_at` with `detected_at.astimezone(...)`. The `detected_at` values loaded from the alerts table are naive datetimes stored in UTC, and `astimezone()` interprets a naive datetime as *local* time. As a result, when `edr monitor` runs on a machine whose timezone is not UTC, the "Time:" field of every alert (and `formatted_detected_at` in the source freshness result description) shows the **UTC value labeled with the local timezone name**.

In the same source freshness message, `max_loaded_at` goes through `convert_datetime_utc_str_to_timezone_str()`, which correctly treats the value as UTC. So a single message mixes two different timezones.

Real-world example (container with `TZ=Asia/Tokyo`, no `timezone` config; the check actually ran at 18:08:22 JST = 09:08:22 UTC):

```
Time: 2026-07-22 09:08:22 JST
When the test ran at 2026-07-22 09:08:22 JST, the most recent record found in
the table was 1 day 0h 11m 41s earlier (2026-07-21 17:55:10 JST).
```

Note the internal inconsistency: `09:08:22 − (17:55:10 the previous day)` is 15h13m, not "1 day 0h 11m 41s". The elapsed time (computed in UTC) and `max_loaded_at` (correctly converted to JST) are right; only the "test ran at" timestamp is mislabeled. This misled our team into believing the alert had been delayed by 9 hours.

**To Reproduce**

Run the following on a machine whose timezone is not UTC (or keep the two `TZ` lines to force it):

```python
import os, time
os.environ["TZ"] = "Asia/Tokyo"; time.tzset()

from datetime import datetime
from elementary.monitor.alerts.source_freshness_alert import SourceFreshnessAlertModel

alert = SourceFreshnessAlertModel(
    id="x", source_name="s", identifier="t", original_status="fail",
    path="p", error=None, alert_class_id="c", source_freshness_execution_id="e",
    detected_at=datetime(2026, 7, 22, 9, 8, 22),          # naive UTC, as loaded from the alerts table
    max_loaded_at=datetime(2026, 7, 21, 8, 55, 10),       # naive UTC
    max_loaded_at_time_ago_in_s=87101.8,
)
print(alert.detected_at_str)      # 2026-07-22 09:08:22 JST  <- UTC value, JST label
print(alert.result_description)   # mixes 09:08:22 "JST" with a correctly converted 17:55:10 JST
```

Setting the `timezone` config does **not** help while the host timezone is non-UTC: the naive value is assumed to be local, so converting local→local is a no-op.

| Host TZ | `timezone` config | "Time:" shown | correct? |
|---|---|---|---|
| Asia/Tokyo | (none) | 09:08:22 JST | ✗ |
| UTC | (none) | 09:08:22 UTC | ✓ |
| UTC | Asia/Tokyo | 18:08:22 JST | ✓ |
| Asia/Tokyo | Asia/Tokyo | 09:08:22 JST | ✗ |

**Expected behavior**

Naive `detected_at` should be interpreted as UTC (matching how it is stored and how `max_loaded_at` is handled in the same message), so the example above renders `Time: 2026-07-22 18:08:22 JST`.

**Screenshots**

**Environment (please complete the following information):**

- Elementary CLI (edr) version: 0.23.4 (the relevant code is unchanged on v0.25.1 and current master)
- Elementary dbt package version: 0.25.1
- dbt version you're using: 1.11.12
- Data warehouse: snowflake
- Infrastructure details: Linux Docker container with `ENV TZ Asia/Tokyo`, running `edr monitor` on AWS ECS Fargate (prod)

**Additional context**

Affects the "Time:" field of all alert types via `AlertModel.detected_at_str`, and additionally `formatted_detected_at` inside the source freshness result description. Deployments whose host timezone is UTC never see the bug, which is probably why it hasn't been reported.

**Would you be willing to contribute a fix for this issue?**

Yes — I have a fix ready (a few lines in `elementary/monitor/alerts/alert.py` interpreting naive `detected_at` as UTC before `astimezone`, plus unit tests) and will open a PR.

zerafachris:
## Problem

Fixes #2304.

`AlertModel.__init__` calls `detected_at.astimezone(...)` on a naive `datetime` value.  Python's [`astimezone()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone) interprets a **naive** datetime as the **local wall-clock time**, not UTC.

Since the alerts table stores timestamps in UTC, running `edr monitor` on a host whose timezone is not UTC produces alerts where the **"Time:" field shows the UTC value labeled with the local timezone name**:

```
# Container with TZ=Asia/Tokyo; test actually ran at 18:08:22 JST (= 09:08:22 UTC)
Time: 2026-07-22 09:08:22 JST   ← UTC value with JST label (wrong)
```

This misled users into thinking alerts were hours delayed.

## Fix

Before calling `astimezone()`, attach `UTC` tzinfo to any naive datetime so the conversion is always relative to UTC:

```python
detected_at_utc_aware = (
    detected_at.replace(tzinfo=dt_timezone.utc)
    if detected_at.tzinfo is None
    else detected_at
)
self.detected_at_utc = detected_at_utc_aware
self.detected_at = detected_at_utc_aware.astimezone(...)
```

`detected_at_utc` is also updated to hold the timezone-aware value for consistency (previously it held the raw naive input).

## Test

Added `tests/unit/alerts/test_alert_detected_at_timezone.py` with three cases:
- Naive UTC → JST: 09:08:22 UTC must become 18:08:22 JST
- Already-aware UTC → JST: no double-conversion
- `None` → fields stay `None`, `detected_at_str == "N/A"`

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected alert timestamp handling so timezone-free timestamps are consistently interpreted as UTC.
  - Preserved correct behavior for timestamps that already include timezone information.
  - Prevented alert filtering errors for alerts with a “reused” status.
  - Alerts marked as reused are now excluded by default status filters.

- **Improvements**
  - Added support for the new “reused” alert status.

### DIFF
--- a/.github/workflows/test-warehouse.yml
+++ b/.github/workflows/test-warehouse.yml
@@ -97,6 +97,17 @@ jobs:
         with:
           path: elementary
           ref: ${{ inputs.elementary-ref }}
+          # Don't leave the base repo's token in the git config: the checked out
+          # code is fork-controlled (see below) and nothing here pushes.
+          persist-credentials: false
+          # Fork PRs are tested via pull_request_target intentionally: the
+          # integration tests need the CI secrets of our test warehouses, which
+          # a pull_request run of a fork cannot access. The caller
+          # (test-all-warehouses.yml) gates fork PRs behind the
+          # elementary_test_env environment, so a maintainer must approve the
+          # run before this fork code is checked out and executed, and only an
+          # explicit allowlist of secrets is passed in.
+          allow-unsafe-pr-checkout: true
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
@@ -110,6 +121,7 @@ jobs:
           repository: elementary-data/dbt-data-reliability
           path: dbt-data-reliability
           ref: ${{ inputs.dbt-data-reliability-ref }}
+          persist-credentials: false
 
       - name: Validate workflow inputs
         run: |
--- a/elementary/monitor/alerts/alert.py
+++ b/elementary/monitor/alerts/alert.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone as dt_timezone
 from typing import Dict, List, Optional
 
 from dateutil import tz
@@ -58,8 +58,16 @@ class AlertModel:
         self.detected_at = None
         if detected_at is not None:
             try:
-                self.detected_at_utc = detected_at
-                self.detected_at = detected_at.astimezone(
+                # Timestamps from the DB are stored as UTC but arrive as naive
+                # datetimes.  Python's astimezone() interprets a naive datetime
+                # as *local* time, so we must attach UTC tzinfo first.
+                detected_at_utc_aware = (
+                    detected_at.replace(tzinfo=dt_timezone.utc)
+                    if detected_at.tzinfo is None
+                    else detected_at
+                )
+                self.detected_at_utc = detected_at_utc_aware
+                self.detected_at = detected_at_utc_aware.astimezone(
                     tz.gettz(timezone) if timezone else tz.tzlocal()
                 )
             except Exception:
--- a/elementary/monitor/data_monitoring/schema.py
+++ b/elementary/monitor/data_monitoring/schema.py
@@ -22,6 +22,7 @@ class Status(str, Enum):
     ERROR = "error"
     RUNTIME_ERROR = "runtime error"
     PARTIAL_SUCCESS = "partial success"
+    REUSED = "reused"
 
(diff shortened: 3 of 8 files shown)


haritamar:
Closing in favor of #2305 which handles the same issue

PR #1940, opened 2025-06-20
Fix BigQuery scan for test result rows

tlangton3:
For my test result rows, this reduced the scan from 26GB to 35.7MB, as the current implementation omits partition pruning.

PR #2351, opened 2026-09-10
perf: prune partitions in the remaining report queries

tlangton3:
## What

The report queries filter their largest tables with `edr_datediff(edr_cast_as_timestamp(col), now, 'day') < days_back`. On BigQuery, a function applied to the column defeats partition pruning, so these scan all history however small `days_back` is.

This is the same defect #1940 fixed for `test_result_rows`, applied to the remaining sites — and it generalises that fix so the pattern is in one place instead of copied per macro.

## Measured

A 7-day lookback selecting a single column, on tables exactly as they are — no repartitioning needed to get this:

| Table | Before | After |
|---|---|---|
| `elementary_test_results` (12M rows, 13GB) | 1,368 MB | **13.68 MB** |
| `dbt_run_results` (17M rows, 41GB) | 1,627 MB | **63.56 MB** |
| `test_result_rows` (26M rows, 41GB) | 1,060 MB | **101 MB** |

For scale, an unfiltered scan of `elementary_test_results` reads 1,279 MB: the current predicate reads *more than no filter at all*, because it also pays to read the column it filters on.

## How

A dispatched `days_back_filter` macro. `default__` renders exactly the predicate these queries have always used, so **nothing changes off BigQuery**. `bigquery__` compares the column directly and adds a bound on the partition column.

Applied to both CTEs of `current_tests_run_results_query`, and to `get_result_rows_agate` — which lets that macro's `bigquery__` override go away. #1940 had to duplicate the entire body just to change one `where` clause; with the dispatch in the predicate helper, the two copies collapse back into one.

### About the redundant-looking bound

The BigQuery variant emits e.g. `detected_at > X and created_at > X-1day`. The second predicate is implied by the first — it exists only because BigQuery prunes on the raw partition column and nothing else. It is needed when the filtered column is not the partition column, and unavoidable for `dbt_run_results`, which stores `generated_at` and `execute_completed_at` as strings: those need a cast, and no cast predicate can ever prune.

It cannot exclude a row the real predicate wants, because `created_at` is stamped by `insert_rows` from `edr_current_timestamp()` at insert time and never read from the row payload — so it is always at or after the event timestamp it guards. Checked over 17.2M rows: `created_at - generated_at` >= 0s, `created_at - execute_completed_at` >= +5s, `created_at - detected_at` >= 0s across 12M and 26M row tables, zero violations of any.

Since those two timestamps are not read from the same clock — one warehouse-side, one client-side — the bound is given a day of slack rather than the same bound, so client drift cannot cost a row. Pruning is at day granularity, so that is at most one extra partition; on the figures above it measured as no difference at all, the real predicate being the binding constraint.

## Not included

`get_models_runs` has the same defect and is the largest single win, but it reads the `model_run_results` view, which does not expose `created_at` — so it has no prunable column until elementary-data/dbt-data-reliability#1057 lands. Deliberately left for a follow-up rather than shipped broken.

`get_source_freshness_results` and `can_upload_source_freshness` carry the same shape. I have no source-freshness data to measure, so I have not touched them — happy to if you would like them included.

## Testing

Rendering verified for each branch of the macro: timestamp column, string column (cast applied), the column being the partition column itself (bound correctly suppressed), a string-typed `days_back`, and the default path (byte-identical semantics to the predicate it replaces). Resulting predicates dry-run against real tables for the figures above. `typos` clean; the repo's Python hooks do not apply to `.sql`.

## Summary by CodeRabbit

- **Performance**
  - Improved date-range filtering for recent test and run results.
  - BigQuery queries can make better use of partitioned data, potentially reducing scan volume and improving response times.

- **Consistency**
  - Standardized result filtering across supported database environments.
  - Preserved filtering by valid result IDs and grouping of returned rows.

### DIFF
--- a/elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
+++ b/elementary/monitor/dbt_project/macros/base_queries/current_tests_run_results_query.sql
@@ -2,14 +2,14 @@
     with elementary_test_results as (
         select * from {{ ref('elementary', 'elementary_test_results') }}
         {% if days_back %}
-            where {{ elementary.edr_datediff(elementary.edr_cast_as_timestamp('detected_at'), elementary.edr_current_timestamp(), 'day') }} < {{ days_back }}
+            where {{ elementary_cli.days_back_filter('detected_at', days_back, partition_column='created_at') }}
         {% endif %}
     ),
 
     dbt_run_results as (
         select * from {{ ref('elementary', 'dbt_run_results') }}
         {% if days_back %}
-            where {{ elementary.edr_datediff(elementary.edr_cast_as_timestamp('execute_completed_at'), elementary.edr_current_timestamp(), 'day') }} < {{ days_back }}
+            where {{ elementary_cli.days_back_filter('execute_completed_at', days_back, partition_column='created_at', column_is_string=true) }}
         {% endif %}
     ),
 
--- a/elementary/monitor/dbt_project/macros/get_result_rows_agate.sql
+++ b/elementary/monitor/dbt_project/macros/get_result_rows_agate.sql
@@ -1,32 +1,10 @@
 {% macro get_result_rows_agate(days_back, valid_ids_query = none) %}
-  {% do return(adapter.dispatch('get_result_rows_agate', 'elementary_cli')(days_back, valid_ids_query)) %}
-{% endmacro %}
-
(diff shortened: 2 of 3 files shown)


tlangton3:
Follow-up that builds on this: #2352 applies the same `days_back_filter` to `get_models_runs`. Left as a draft because it additionally needs elementary-data/dbt-data-reliability#1057 to expose `created_at` on the `model_run_results` view.

PR #2360, opened 2026-09-22
Remove the CLI's hard dbt-core dependency (dbt v2 ships as dbt/dbt-oss)

haritamar:
## Summary

Removes the CLI's hard dependency on `dbt-core`, now that dbt v2 ships as the `dbt` / `dbt-oss` PyPI packages while `dbt-core` stays on 1.x (CORE-1477, follow-up to #2333).

- **pyproject**: drop the top-level `dbt-core = ">=1.8,<3.0.0"` requirement. dbt-core 1.x comes through the adapter extras, and dbt v2 needs no Python dependency at all (`pip install elementary-data` + `pip install dbt`/`dbt-oss`).
- **Extras**: add `dbt1-<warehouse>` aliases (e.g. `dbt1-snowflake`) for every existing extra, plus `dbt1-all`. The new names are the ones the docs will use; the old names remain supported indefinitely. All extras (new and legacy) now also explicitly include `dbt-core >=1.8,<2.0.0`, so they can never resolve a dbt-core 2.x release (the adapters themselves only lower-bound dbt-core).
- **Detection** (`dbt_installation.py`): `get_dbt_package_version()` → `get_dbt2_package_version()`, which checks both the `dbt` and `dbt-oss` distributions and only counts versions `>= 2` (a legacy 1.x `dbt` metapackage doesn't qualify).
- **Factory** (`get_dbt_runner_method()`):
  - `RunnerMethod.API`/`SUBPROCESS` renamed to `DBT1_API`/`DBT1_SUBPROCESS` to make it clear they drive dbt-core 1.x; the old names remain as enum aliases (same values), so existing callers and `DBT_RUNNER_METHOD=api|subprocess` keep working.
  - When dbt-core 1.x and a dbt v2 installation coexist in one env, **dbt v2 now takes precedence** (per review), with a log explaining how to force dbt-core (`DBT_RUNNER_METHOD=api`).
  - When neither dbt-core nor a dbt v2 installation is detected but a `dbt` executable is on PATH (e.g. system-wide/pipx dbt 1.x without pip metadata in the venv), fall back to `DBT1_SUBPROCESS` as before.
  - When nothing is found at all, raise the new `NoDbtInstallationError` with actionable install instructions instead of failing later with an obscure subprocess error.

No behavior change for the documented install paths: `elementary-data[<warehouse>]` users get the exact same resolution as today (unless they also install dbt v2, which now wins).

CI fixes (unrelated breakage that surfaced on this PR):
- MinIO images (trino/dremio e2e, incl. the hard-coded `mc` image in `external_seeders/dremio.py`) moved from Docker Hub to quay.io — the Docker Hub repos became unavailable.
- Bumped the CI clickhouse-server image 24.3 → 24.8 (LTS): dbt-clickhouse 1.10.3 sends `lightweight_deletes_sync=3` as a default connection setting, which is `UNKNOWN_SETTING` on clickhouse-server < 24.5, failing on the first `CREATE DATABASE`. Verified 1.10.3 works against 24.8 locally and in CI.
- Spark job: re-assume the AWS OIDC role after the (slow, up to ~50 min) Spark docker image build — the default 1-hour session was expiring before the end-of-job S3 report upload.

A separate docs PR (base `docs`, #2361) restructures the install instructions around the two paths (dbt v2: no extra; dbt-core 1.x: `dbt1-<warehouse>`).

## Testing

- Unit tests extended: no-installation error, metadata-free PATH fallback, dbt v2 precedence on coexistence, `dbt`/`dbt-oss`/1.x-metapackage detection matrix, legacy enum alias identity. Full unit suite: 488 passed.
- Wheel built locally and its METADATA inspected to confirm the extras/aliases and absence of a top-level dbt requirement.
- dbt-clickhouse root cause reproduced locally: `CREATE DATABASE` with `lightweight_deletes_sync` fails on clickhouse-server 24.3 (`Code: 115 UNKNOWN_SETTING`); succeeds on 24.8.
- mypy + pre-commit (black/isort/flake8) clean.
- Full warehouse CI matrix green (all 1.x warehouses + fusion/dbt2 targets).

Link to Devin session: https://app.devin.ai/sessions/496f2bf70ae14ec1a1bccb245fbc9328
Open in Devin Desktop: https://app.devin.ai/desktop/session/496f2bf70ae14ec1a1bccb245fbc9328?variant=devin
Requested by: @haritamar


elementary/clients/dbt/factory.py
```diff
@@ -60,14 +65,26 @@ def get_dbt_runner_method() -> RunnerMethod:
     if dbt_core_version is not None:
         if dbt_core_version.major >= 2:
```
haritamar: This maybe deserves a small comment (1-2 lines) - I think the reason to keep it is that dbt-core has some 2.X pre-releases.
Not sure if they plan to remove them, but in case they don't we should keep this check.

elementary/clients/dbt/factory.py
```diff
@@ -60,14 +65,26 @@ def get_dbt_runner_method() -> RunnerMethod:
...
             return RunnerMethod.DBT2
+        if is_dbt2_binary_available():
+            logger.info(
+                f"Both dbt-core {dbt_core_version} and a dbt v2 installation were "
+                "detected - using dbt-core. To use dbt v2 instead, uninstall "
+                "dbt-core (and the elementary-data adapter extra) or set "
+                "DBT_RUNNER_METHOD=dbt2."
```
haritamar: I think maybe now that dbt2 is GA, it should actually take precedence over dbt-core.

elementary/clients/dbt/factory.py
```diff
@@ -60,14 +65,26 @@ def get_dbt_runner_method() -> RunnerMethod:
...
         return RunnerMethod.SUBPROCESS
 
     if is_dbt2_binary_available():
         return RunnerMethod.DBT2
 
-    return RunnerMethod.SUBPROCESS
+    # dbt may be installed without pip package metadata (e.g. a system-wide or
+    # pipx-managed dbt 1.x) - fall back to running it as a subprocess.
+    if shutil.which("dbt"):
+        return RunnerMethod.SUBPROCESS
```
haritamar: I'm wondering if to be clear we should rename the dbt1 methods to `DBT1_SUBPROCESS` and `DBT1_API`
(and probably keep the old ones as legacy aliases, but at least in the code refer to the newer ones).

WDYT? I feel it's a bit more clear going forward
as dbt1 is not clearly the default anymore.

elementary/clients/dbt/factory.py
haritamar, line 68: I'm not sure if we want to do a small rename here too (and below), and replace the term "dbt_core" with "dbt1", wdyt?

elementary/clients/dbt/dbt_installation.py
```diff
@@ -21,22 +21,26 @@ def get_dbt_core_version() -> Optional[version.Version]:
...
 def is_dbt2_binary_available() -> bool:
     env_path = os.getenv(DBT_FUSION_PATH_ENV_VAR)
     if env_path and os.path.exists(os.path.expanduser(env_path)):
         return True
 
-    dbt_package_version = get_dbt_package_version()
-    if dbt_package_version is not None and dbt_package_version.major >= 2:
+    if get_dbt2_package_version() is not None:
         return True
     return os.path.exists(os.path.expanduser(DEFAULT_DBT_FUSION_PATH))
```
haritamar, line 45: Just to make sure - is this check still reliable to verify dbt2 installation?

